PR #37, opened 2026-05-04
ci(quality.yml): build frontend + PHP-server health-check before Playwright

rubenvdlinde:
## Summary

The Playwright job had two latent issues that surfaced when mydash recently flipped its \`enable-playwright\` gate:

1. **Missing frontend build.** Vue / React apps that mount via \`Util::addScript()\` need their webpack bundle on disk by the time specs render the page. The job ran \`npm ci\` + \`playwright install\` but never \`npm run build\` — the rendered page loaded a 404 script tag and the SPA never mounted. Every selector wait timed out on the page's first hydration marker.
2. **Backgrounded PHP server fragility.** \`Start PHP built-in server\` does \`php -S 0.0.0.0:8080 &\` then ends. The orphaned process is normally reparented to init and survives, but subsequent multi-minute steps sometimes leave it dead by the time specs run.

## Changes

**New step: \`Build app frontend\`** (between Install Playwright and Validate Playwright tests exist)
- Probes \`package.json\` for an \`npm run build\` script via \`node -e\`. Skips silently when missing so legacy apps without a frontend keep working.
- Runs \`npm run build\` once on a fresh runner.

**New step: \`Verify PHP server still alive (restart if dead)\`** (right before Run Playwright tests)
- \`curl -sf /status.php\` health check.
- On failure: \`pkill\` any stale \`php -S\` process, re-start via \`nohup … & disown\`, wait for next /status.php to return 200.
- Logs both the alive and the restart branch so the action log is unambiguous.

## Test plan
- [ ] mydash will flip \`enable-playwright: true\` once this PR merges; that flip is the canonical end-to-end test for both fixes.
- [x] Existing apps without a build script (legacy widgets, PHP-only) continue to work — the build step's \`node -e\` probe returns falsy and skips.

## Context
mydash's parallel commit (`fd0cdb5`) added a per-app `tests/e2e/global-setup.ts` `ensureBundleBuilt()` helper that does the same build via Playwright globalSetup. Once this PR merges and the gate flips green, that helper becomes a no-op (bundle already on disk by the time globalSetup runs) — kept for local dev where the workflow doesn't run.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -1081,6 +1081,21 @@ jobs:
           npm install --save-dev @playwright/test
           npx playwright install --with-deps chromium
 
+      - name: Build app frontend
+        # Vue/React apps that mount via `Util::addScript()` need their
+        # webpack bundle on disk before specs render the page; without
+        # this step every selector waits times out because the script
+        # tag 404s and the SPA never mounts. Skip silently when the app
+        # has no `build` npm script (legacy apps without a frontend).
+        run: |
+          cd server/apps/${{ inputs.app-name }}
+          if [ -f "package.json" ] && node -e "process.exit(require('./package.json').scripts && require('./package.json').scripts.build ? 0 : 1)" 2>/dev/null; then
+            echo "Building app frontend with 'npm run build'…"
+            npm run build
+          else
+            echo "No 'build' script in package.json; skipping frontend build."
+          fi
+
       - name: Validate Playwright tests exist
         run: |
           cd server/apps/${{ inputs.app-name }}
@@ -1095,6 +1110,32 @@ jobs:
             exit 1
           fi
           echo "Found $TESTS Playwright test file(s) in $TEST_PATH."
+
+      - name: Verify PHP server still alive (restart if dead)
+        # The "Start PHP built-in server" step backgrounds `php -S` and
+        # exits; subsequent steps run in fresh shells. The orphaned
+        # process is normally reparented to init and survives, but on
+        # busy CI runners we've seen it gone by the time specs run
+        # (especially after a multi-minute frontend build). Probe for a
+        # live server here and re-start when needed so spec-phase
+        # navigations always reach a working backend.
+        run: |
+          cd server
+          if ! curl -sf -o /dev/null -m 5 http://localhost:8080/status.php; then
+            echo "::warning::PHP server not reachable at :8080 — re-starting."
+            # Kill any stale `php -S` processes that are stuck on the
+            # port without serving requests.
+            pkill -f "php -S 0.0.0.0:8080" || true
+            sleep 1
+            nohup php -S 0.0.0.0:8080 > /tmp/php-server.log 2>&1 &
+            disown
+            sleep 3
+            curl -sf -o /dev/null -m 5 http://localhost:8080/status.php \
+              || (echo "::error::PHP server still not reachable after re-start." && cat /tmp/php-server.log && exit 1)
+            echo "PHP server re-started successfully."
+          else
+            echo "PHP server is alive."
+          fi
 
       - name: Run Playwright tests
         run: |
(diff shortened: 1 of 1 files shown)


rubenvdlinde:
Goede feedback, doorgevoerd in de laatste push (3622563):

**Restructuring i.p.v. patchen** — de PHP-server start nu pas ná de trage setup-fase (npm ci + Playwright install + frontend build), zodat het venster tussen 'start' en 'eerste spec' seconden is i.p.v. minuten. Daarmee is het hele "Verify PHP server still alive (restart if dead)" stuk weg — geen `pkill -f` meer (dat was inderdaad een kwetsbaarheid voor unintended matches).

Nieuwe volgorde (precies zoals voorgesteld):
1. Checkout / setup PHP / install Nextcloud / install app
2. Setup Node / install npm deps / install Playwright
3. Build app frontend
4. Validate Playwright tests exist
5. Start PHP built-in server  ← hierheen verplaatst
6. Seed test data             ← hierheen verplaatst
7. Run Playwright tests

Server-start logt nu z'n eigen PID naar `/tmp/php-server.pid` en faalt-fast met `exit 1` + log dump als de probe na 3s niet reageert; geen guardrail meer nodig dus.

Stats: -36 / +21 lines (de hele restart-met-`pkill` step is verdwenen).

> zit btw geen branch protection op die repo?

Nee, klopt — die zou er moeten komen. Buiten scope voor deze PR maar wel een terechte vlag. Per memory zijn er org-rulesets voor `dev`/`beta`/`main` op andere repos; `.github` is daar expliciet van uitgesloten omdat deze repo voor workflow-updates juist zonder review moet kunnen. Misschien een aparte issue waard om dat opnieuw te bekijken — een 1-review minimum lijkt op deze repo zinvoller dan helemaal niks gegeven de blast radius.

.github/workflows/quality.yml
```diff
@@ -1096,6 +1096,32 @@ jobs:
...
+        # Started AFTER the slow setup phases (npm ci, playwright install,
+        # frontend build) so the server is not idle for minutes — that's
+        # what was causing the orphaned `php -S` process to disappear by
+        # the time specs ran. Probe right after start and fail fast if
+        # the server is not actually listening.
+        run: |
+          cd server
+          php -S 0.0.0.0:8080 > /tmp/php-server.log 2>&1 &
+          echo $! > /tmp/php-server.pid
+          sleep 3
```
WilcoLouwerse: 🟡 **Fixed `sleep 3` + single curl probe is fragile under runner load.**

`curl -sf` returns failure immediately on connection refused — the `-m 5` timeout only applies after the connection is established. So if `php -S` hasn't bound to `:8080` by the time `sleep 3` finishes (rare, but possible on a noisy runner), the step fails spuriously without a retry, with no diagnostic beyond "failed to come up".

A polling loop with a hard timeout would make the new fail-fast path more honest:

```bash
php -S 0.0.0.0:8080 > /tmp/php-server.log 2>&1 &
PID=$!
timeout 15 bash -c 'until curl -sf -o /dev/null http://localhost:8080/status.php; do sleep 0.5; done' || {
  echo "::error::PHP built-in server failed to come up on :8080."
  cat /tmp/php-server.log
  exit 1
}
echo "PHP built-in server alive (pid=$PID)."
```

Not a blocker — `php -S` typically binds in well under 100 ms — but worth tightening since the whole point of moving the server start next to the test step was to make the failure mode less spooky.

WilcoLouwerse (approved):
No blockers. One concern noted inline: [fixed `sleep 3` before single curl probe](https://github.com/ConductionNL/.github/pull/37#discussion_r3187054339) — `curl -sf` fails immediately on connection refused, so a polling loop would make the new fail-fast path more honest. The reorder is the right call (root-cause fix instead of `pkill` guardrail), and the `node -e` build-script probe cleanly preserves legacy-app compatibility.

Heads-up: PR is currently in `CONFLICTING` state vs `main` — needs a rebase before merge.

.github/workflows/quality.yml
rubenvdlinde: Doorgevoerd in 5530d7c — exact je voorgestelde `timeout 15` + `until curl … sleep 0.5` patroon overgenomen. Bind-variance is nu opgevangen, fail-fast pad blijft eerlijk over wat er fout ging.

PR #60, opened 2026-05-12
feat(global-settings): v1.7.0 — kernel-level immutability + tool-write hook

WilcoLouwerse:
## Summary

Upgrades the global Claude settings to v1.7.0 with a four-layer security model:

- **Layer 1** (existing): deny-list in `settings.json`
- **Layer 2** (enhanced): `block-write-commands.sh` — adds `chattr` guard and script-body scanner that reads invoked script files and checks their bodies for protected-path write operations
- **Layer 3** (NEW): `block-config-tool-writes.sh` — `PreToolUse` hook for `Write|Edit|MultiEdit` calls; denies direct writes to `~/.claude/` config files AND script content that would write to them when executed
- **Layer 4** (NEW): `chattr +i` kernel immutability — README install + update flow use `chattr`, replacing the old `chmod` re-lock approach

### Key fixes in this implementation
- Tilde-expansion bug in `${var#~/}` pattern: bash tilde-expands `~/` inside parameter substitutions before matching — fixed by quoting as `'~'/` in all affected spots
- Python `os.system('rm ~/.claude/...')` bypass: guard-2 regex expanded to catch commands inside shell string arguments `('` and `"`)

## Test plan

- [x] `tests/test-block-write-commands.sh` — 7449 tests, 100% pass (covers chattr guard, script-body scanner, push-auth, all write vectors)
- [x] `tests/test-block-config-tool-writes.sh` — 65 tests, 100% pass (covers all tool-write guard paths including tilde paths and Python content scan)

## Install notes (for existing users)

See README.md → Updating section. The `chattr +i` lock requires `sudo` to clear before updating and re-apply after. The `block-config-tool-writes.sh` hook is new — copy it to `~/.claude/hooks/` during the update.


rjzondervan (approved):

rubenvdlinde (approved):
**Review summary (consolidation triage pass):**

- Substantive defence-in-depth: 4-layer model (deny-list / Bash hook with chattr+script-body scan / new tool hook for Write|Edit|MultiEdit / kernel `chattr +i`). Overlap is intentional and explicitly justified in the README.
- Concrete bypasses addressed: tilde-expansion in `${var#~/}`, Python `os.system('rm ~/.claude/...')` content scan.
- Test coverage: 7449 + 195 lines of tests claimed passing.
- ShellCheck pass, both `test` jobs pass.
- CodeQL JS failure on this PR is a transient `429 Too Many Requests` from codeload.github.com fetching codeql-action — infra flake, will rerun.
- Migration cost noted: existing users need `sudo chattr -i` before update + `sudo chattr +i` after. README documents both steps.

Approving.

PR #116, opened 2026-07-31
ci(quality): enable-php switch, custom frontend checks, strict npm ci, and a real Quality Report gate

remko48:
## What

Four improvements to the shared quality workflow, aimed at making `quality / Quality Report` usable as the single org-wide required status check and letting every repo — PHP app or JS-only — adopt the same wrapper.

### 1. `enable-php` master switch
New boolean input (default `true`) that gates the entire PHP toolchain: `php-quality`, PHPUnit, the composer legs of security/license, and SBOM generation. JS-only repos (`nextcloud-vue`, `conduction-theme`) can now adopt the shared workflow with `enable-php: false` — the skipped PHP checks still report as "skipped", which satisfies required status checks, so the same org-wide required contexts work for every repo.

### 2. Custom frontend checks
- **`frontend-checks`** — JSON array of npm scripts, each running as its own `Frontend Check (<script>)` matrix job (fresh checkout + `npm ci` per leg).
- **`frontend-setup-command`** — optional additional setup that runs *after* the root `npm ci` (never replaces it), e.g. installing a nested package's dependencies.
- **`node-version`** — replaces the 7 hardcoded Node `"20"` pins (default unchanged).
- Listed scripts are validated against `package.json` **before** install — a missing script fails fast with an actionable error annotation pointing to CONVENTIONS.md.

### 3. Quality Report becomes a real gate
The `report` job previously always succeeded (`if: always()` + report-only steps), so requiring it in branch protection gated nothing. It now:
- fails when any upstream job failed (new gate step), and
- covers the complete job set — `needs:` extended with `frontend-checks`, `sbom`, `features-check`, `features-extract`, so e.g. a stale `docs/features.json` on a PR trips the gate.

### 4. Strict `npm ci` (removal of `--legacy-peer-deps`)
The hardcoded `--legacy-peer-deps` on all `npm ci` invocations is removed. Peer-resolution behaviour is now **repo-owned** via a committed `.npmrc` (the pattern openconnector already uses), so CI behaves exactly like local/production installs and real peer conflicts surface instead of being masked fleet-wide. CONVENTIONS.md documents the policy: the flag is an emergency brake for blocked releases only — committed in `.npmrc` with a dated why-comment and a linked tracking issue, removed ASAP.

> ⚠️ Rollout note: repos whose lockfiles require legacy resolution but lack the `.npmrc` line will fail their npm jobs on the first run after this merges — that's the conflict surfacing, not a workflow bug. openregister was verified to pass strict `npm ci` cleanly.

## Security hardening

All caller-supplied values used by the new `frontend-checks` job (`matrix.script`, `frontend-setup-command`) are passed through `env:` and referenced as `"$VAR"` instead of being inlined into `run:` scripts, per GitHub's security-hardening guidance.

## Documentation

`CONVENTIONS.md` gains three sections under the Central Quality workflow: **JS-only repos (`enable-php: false`)**, **Custom frontend checks** (script requirements: must exist, must be self-contained, bundle order-dependent steps), and the **peer-dependency policy**.

## Review

All four change requests from @WilcoLouwerse are addressed: `sbom` gated via `enable-php` (option 1), report `needs:` completed, env-var interpolation hardening, and input descriptions now state the `package.json` + `package-lock.json` prerequisites. The suggested hardening pass for the pre-existing `newman-seed-command` / `playwright-seed-command` / `journeydoc-*` inputs is left as a follow-up.

## Compatibility

All new inputs default to current behaviour (`enable-php: true`, Node 20, empty checks list). The two intentional behavioural changes for existing adopters: Quality Report can now be red, and `npm ci` is strict.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -22,6 +22,11 @@ on:
         required: false
         type: string
         default: '["stable31", "stable32"]'
+      enable-php:
+        description: "Master switch for the PHP toolchain (php-quality, PHPUnit, composer security/license legs). Set false for JS-only repos without composer.json."
+        required: false
+        type: boolean
+        default: true
       enable-psalm:
         description: "Run Psalm static analysis"
         required: false
@@ -57,6 +62,21 @@ on:
         required: false
         type: boolean
         default: true
+      node-version:
+        description: "Node.js version for all frontend/npm steps"
+        required: false
+        type: string
+        default: "20"
+      frontend-setup-command:
+        description: "Optional extra install command run after `npm ci` in the frontend-checks job (e.g. 'cd docusaurus && npm ci')"
+        required: false
+        type: string
+        default: ""
+      frontend-checks:
+        description: "JSON array of npm scripts to run as individual 'Frontend Check (<script>)' jobs (e.g. '[\"test\", \"check:docs\"]'). Requires enable-frontend."
+        required: false
+        type: string
+        default: "[]"
       enable-license-check:
         description: "Run dependency license compliance check (composer + npm)"
         required: false
@@ -183,6 +203,7 @@ jobs:
 
   # ── Group 1: PHP Quality ────────────────────────
   php-quality:
+    if: ${{ inputs.enable-php }}
     runs-on: ubuntu-latest
     name: "PHP Quality (${{ matrix.tool }})"
     strategy:
@@ -293,7 +314,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "20"
+          node-version: "${{ inputs.node-version }}"
           cache: "npm"
 
       - name: Install dependencies
@@ -326,6 +347,55 @@ jobs:
           name: result-vue-quality-${{ matrix.tool }}
           path: quality-results/
 
+  # ── Group 2b: Custom frontend checks ───────────
+  #
+  # Repo-defined npm scripts run as individual matrix legs, e.g.
+  #   frontend-checks: '["test", "check:build", "check:docs"]'
+  # Each leg gets a fresh checkout + npm ci, so scripts must be
+  # self-contained (bundle ordering deps into one script, e.g.
+  # "check:build": "npm run build && npm run check:css-entry").
+  frontend-checks:
+    if: ${{ inputs.enable-frontend && inputs.frontend-checks != '[]' }}
+    runs-on: ubuntu-latest
+    name: "Frontend Check (${{ matrix.script }})"
+    strategy:
+      matrix:
+        script: ${{ fromJSON(inputs.frontend-checks) }}
+      fail-fast: false
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "${{ inputs.node-version }}"
+          cache: "npm"
+
+      - name: Install dependencies
+        run: npm ci --legacy-peer-deps
+
+      - name: Extra setup
+        if: ${{ inputs.frontend-setup-command != '' }}
+        run: ${{ inputs.frontend-setup-command }}
+
+      - name: Run ${{ matrix.script }}
+        run: npm run "${{ matrix.script }}"
+
+      - name: Record result
+        if: always()
+        run: |
+          mkdir -p quality-results
+          # ':' is invalid in artifact file paths — sanitize the script name
+          echo "${{ job.status }}" > "quality-results/$(echo '${{ matrix.script }}' | tr ':/' '--').txt"
+      - name: Upload result
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: result-frontend-check-${{ strategy.job-index }}
+          path: quality-results/
+
   # ── Group 3: Security ──────────────────────────
   security:
     runs-on: ubuntu-latest
@@ -350,13 +420,17 @@ jobs:
         if: matrix.ecosystem == 'npm'
         uses: actions/setup-node@v4
         with:
-          node-version: "20"
+          node-version: "${{ inputs.node-version }}"
           cache: "npm"
 
       - name: Run ${{ matrix.ecosystem }} audit
         run: |
           case "${{ matrix.ecosystem }}" in
             composer)
+              if [ "${{ inputs.enable-php }}" != "true" ]; then
+                echo "PHP toolchain is disabled — skipping composer audit."
+                exit 0
+              fi
               composer install --no-progress --prefer-dist --optimize-autoloader
               composer audit --format=plain --abandoned=report
               ;;
@@ -459,12 +533,16 @@ jobs:
         if: matrix.ecosystem == 'npm'
         uses: actions/setup-node@v4
         with:
-          node-version: "20"
+          node-version: "${{ inputs.node-version }}"
           cache: "npm"
 
       - name: Install dependencies
         run: |
           if [ "${{ matrix.ecosystem }}" = "composer" ]; then
+            if [ "${{ inputs.enable-php }}" != "true" ]; then
+              echo "PHP toolchain is disabled — skipping composer install."
+              exit 0
+            fi
             composer install --no-progress --prefer-dist --optimize-autoloader
           else
             npm ci --legacy-peer-deps
@@ -474,6 +552,11 @@ jobs:
       - name: Check ${{ matrix.ecosystem }} licenses
         run: |
           set -euo pipefail
+
+          if [ "${{ matrix.ecosystem }}" = "composer" ] && [ "${{ inputs.enable-php }}" != "true" ]; then
+            echo "PHP toolchain is disabled — skipping composer license check."
+            exit 0
+          fi
 
           # ── Default allowlist (EUPL-1.2 / AGPL-3.0 compatible) ──
           DEFAULT_ALLOWLIST='[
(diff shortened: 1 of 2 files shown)


rjzondervan (approved):

SudoThijn (approved):

.github/workflows/quality.yml
```diff
@@ -22,6 +22,11 @@ on:
         required: false
         type: string
         default: '["stable31", "stable32"]'
+      enable-php:
+        description: "Master switch for the PHP toolchain (php-quality, PHPUnit, composer security/license legs). Set false for JS-only repos without composer.json."
```
WilcoLouwerse: 🟡 **Concern — `enable-php` does not gate the `sbom` job**

The docstring describes this as the "Master switch for the PHP toolchain (php-quality, PHPUnit, composer security/license legs)", but the `sbom` job (line 1970) is not gated by `inputs.enable-php`. `sbom` defaults to `enable-sbom: true` and runs `composer install --no-progress --prefer-dist --optimize-autoloader` unconditionally on pushes to `main` / `beta` / `development`.

For the stated PR goal — "JS-only repos (nextcloud-vue, conduction-theme) can adopt the shared workflow" — this means the caller has to *also* set `enable-sbom: false`, which isn't obvious from either the docstring or the PR body.

Two options:
1. Gate the whole `sbom` job with `enable-php` too (mirrors the `phpunit` pattern at line 760):
   ```yaml
   sbom:
     if: ${{ inputs.enable-php && inputs.enable-sbom && !cancelled() && (...) }}
   ```
2. Extend this description to explicitly call out that `enable-sbom` must also be disabled for JS-only repos:
   > "Master switch for the PHP toolchain (php-quality, PHPUnit, composer security/license legs). Set false for JS-only repos without composer.json. Note: also set `enable-sbom: false` for JS-only repos."

Option 1 keeps the abstraction complete; option 2 preserves the escape hatch of "JS-only repo that still wants an SBOM" (arguably rare).

.github/workflows/quality.yml
```diff
@@ -1649,7 +1732,7 @@ jobs:
   report:
     runs-on: ubuntu-latest
     name: "Quality Report"
-    needs: [php-quality, vue-quality, security, license, phpunit, newman, playwright, journeydoc-capture, baseline-protection]
+    needs: [php-quality, vue-quality, frontend-checks, security, license, phpunit, newman, playwright, journeydoc-capture, baseline-protection]
```
WilcoLouwerse: 🟡 **Concern — "Quality Report" gate misses `features-check`, `sbom`, `features-extract`**

The new gate step below is a real improvement, but it only trips on jobs listed in `needs:`. This list omits three jobs that already exist in the workflow:

- `features-check` — runs on `pull_request` events (line 2121) and can fail if `docs/features.json` is stale. A stale-features PR would produce a red workflow run but a green "Quality Report" check — the exact scenario this PR is trying to prevent.
- `sbom` — only relevant on `main` / `beta` / `development`, but its failure similarly wouldn't trip the gate.
- `features-extract` — non-PR event, but same story.

Suggestion: add at least `features-check` to `needs:` so the PR-time gate is complete. `sbom` and `features-extract` can be added too — since they only run on branch pushes and `report` uses `if: always()`, adding them wouldn't slow down PR runs.

.github/workflows/quality.yml
```diff
@@ -326,6 +347,55 @@ jobs:
...
+        with:
+          node-version: "${{ inputs.node-version }}"
+          cache: "npm"
+
+      - name: Install dependencies
+        run: npm ci --legacy-peer-deps
+
+      - name: Extra setup
+        if: ${{ inputs.frontend-setup-command != '' }}
+        run: ${{ inputs.frontend-setup-command }}
```
WilcoLouwerse: 🟢 **Minor — caller-supplied input inlined into `run:` (defense-in-depth)**

This line and the `Run ${{ matrix.script }}` step below (`npm run "${{ matrix.script }}"`) inline caller-provided inputs directly into the shell. [GitHub's security-hardening guidance](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable) recommends passing untrusted inputs through `env:` and referencing them as `"$VAR"` so that special characters (`"`, backticks, `$()`) can't break out of the intended context.

For a `workflow_call` reusable workflow, the trust boundary is caller-repo maintainers (already trusted), and other places in this file (`newman-seed-command`, `playwright-seed-command`, `journeydoc-*`) use the same inline pattern, so this is not a new regression. Flagging under Strict because the safer pattern is cheap:

```yaml
- name: Extra setup
  if: ${{ inputs.frontend-setup-command != '' }}
  env:
    SETUP_CMD: ${{ inputs.frontend-setup-command }}
  run: bash -c "$SETUP_CMD"

- name: Run frontend check
  env:
    SCRIPT: ${{ matrix.script }}
  run: npm run "$SCRIPT"
```

Optional; can be a follow-up. Same treatment for the pre-existing seed-command inputs would be a nice hardening pass.

.github/workflows/quality.yml
```diff
@@ -57,6 +62,21 @@ on:
...
+        required: false
+        type: string
+        default: "20"
+      frontend-setup-command:
+        description: "Optional extra install command run after `npm ci` in the frontend-checks job (e.g. 'cd docusaurus && npm ci')"
+        required: false
+        type: string
+        default: ""
+      frontend-checks:
+        description: "JSON array of npm scripts to run as individual 'Frontend Check (<script>)' jobs (e.g. '[\"test\", \"check:docs\"]'). Requires enable-frontend."
```
WilcoLouwerse: 🟢 **Minor — description understates prerequisites**

"Requires enable-frontend" is true but the frontend-checks job also unconditionally runs `npm ci --legacy-peer-deps` (line 377), which needs `package.json` **and** `package-lock.json` in the caller repo. Since the PR is aimed at expanding coverage to JS-only repos, worth being explicit so someone turning this on without a lockfile gets a clear reason. Suggestion:

> "JSON array of npm scripts to run as individual 'Frontend Check (<script>)' jobs (e.g. '[\"test\", \"check:docs\"]'). Requires enable-frontend and a repo with package.json + package-lock.json."

Same for `frontend-setup-command` at line 71 — its example (`cd docusaurus && npm ci`) implicitly assumes a nested lockfile.

WilcoLouwerse (changes requested):
## Verdict: REQUEST_CHANGES (Strict)

Good direction — the `enable-php` gate is applied consistently across `php-quality`, `phpunit`, `security` (composer leg), and `license` (composer leg), and the report-gate step is exactly the missing piece that makes "quality / Quality Report" a viable required check.

Two 🟡 gaps prevent the PR from fully delivering its stated goal, plus two 🟢 defense-in-depth notes:

- 🟡 [`enable-php` does not gate `sbom`](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722651) — JS-only repos still fail on `composer install`.
- 🟡 [Quality Report `needs:` omits `sbom` / `features-check` / `features-extract`](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722661) — the "single required check" claim doesn't fully hold.
- 🟢 [Caller-input interpolated directly into `run:`](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722666) — defense in depth.
- 🟢 [`frontend-checks` description understates prerequisites](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722671).

Happy to re-review once addressed.

SudoThijn (approved):

.github/workflows/quality.yml
WilcoLouwerse: Fixed in b08f47f — `sbom.if` now includes `inputs.enable-php` (line 1989), and the docstring lists SBOM alongside the other PHP legs. Option 1 from the original comment.

.github/workflows/quality.yml
WilcoLouwerse: Fixed in b08f47f — `report.needs:` now includes `sbom, features-check, features-extract` (line 1753). Since `report` uses `if: always()` and skipped jobs get `result: skipped` (not `failure`), the extra deps don't slow down PR runs — they only trip the gate when a non-PR-only job actually fails.

.github/workflows/quality.yml
WilcoLouwerse: Fixed in b08f47f — the `Extra setup`, `Run <script>`, and `Record result` steps all use the `env: … "$VAR"` pattern now. Bonus: the new `Verify script exists in package.json` pre-flight step also uses the env-var pattern (`process.env.SCRIPT` inside the node inline) and gives a clear, actionable error if a caller lists a missing script.

.github/workflows/quality.yml
WilcoLouwerse: Fixed in b08f47f — both `frontend-checks` and `frontend-setup-command` descriptions now spell out the `package.json` + `package-lock.json` requirement and link to `CONVENTIONS.md § Custom frontend checks`. The CONVENTIONS.md section itself covers self-contained scripts with a realistic `check:build` example.

WilcoLouwerse (approved):
## Verdict: APPROVE (Standard re-review)

All four prior findings addressed cleanly on `b08f47f`. CI is green (CodeQL + 3 Analyze legs).

- 🟡 [`enable-php` gates `sbom`](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722651) — `sbom.if` updated at line 1989; docstring lists SBOM alongside the other PHP legs.
- 🟡 [Quality Report `needs:` complete](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722661) — `sbom, features-check, features-extract` added at line 1753.
- 🟢 [Env-var pattern applied](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722666) — `Extra setup` / `Run <script>` / `Record result` all switched to `env: … "$VAR"`; bonus pre-flight `Verify script exists in package.json` step.
- 🟢 [Descriptions accurate](https://github.com/ConductionNL/.github/pull/116#discussion_r3690722671) — prerequisites now explicit + linked to `CONVENTIONS.md § Custom frontend checks`.

**Scope observation (not a blocker):** the same PR drops hardcoded `--legacy-peer-deps` from all 8 `npm ci` call sites and adds a substantial "Peer-dependency policy" + "`legacy-peer-deps` is an emergency brake, not a setting" section to `CONVENTIONS.md`. That is a fleet-wide behaviour change — repos with unresolved peer conflicts will start failing on their next quality run and will need to either fix the underlying incompatibility or opt in via `legacy-peer-deps=true` in their own committed `.npmrc`. Worth signalling to the affected repo owners ahead of the merge so they aren't surprised by a red workflow.

Nice cleanup pass — the CONVENTIONS.md addition turns the workflow inputs into properly-documented policy. LGTM.

rubenvdlinde (approved):

PR #126, opened 2026-08-03
ci(quality): add enable-npm and frontend-path inputs, optional Stylelint leg

remko48:
## Summary

Three new inputs for the reusable quality workflow, with matching `CONVENTIONS.md` documentation, covering repo shapes the workflow couldn't serve yet: PHP-only repos and repos with a nested frontend.

### `enable-npm` (default `true`)
Master switch for the npm ecosystem legs of the security and license jobs — the mirror image of `enable-php`. PHP-only repos without a `package.json` (e.g. `openklant`) set it to `false` so those legs skip cleanly instead of failing hard on `npm ci`. Skipped legs still report, so they satisfy the same org-wide required status checks.

### `frontend-path` (default `"."`)
Points every npm-side job (Vue Quality, Frontend Checks, npm security/license legs) at a subdirectory containing `package.json` + `package-lock.json`, for repos where the frontend is not at the repo root (e.g. `woo-website-template-apiv2` with its app in `pwa/`). Applied consistently via:
- job-level `working-directory` defaults,
- `cache-dependency-path` on all `setup-node` steps,
- artifact upload paths,
- `license-checker --start` for the npm license scan.

### `enable-stylelint` (default `true`)
Skips the Stylelint leg of Vue Quality for repos without a `stylelint` npm script, mirroring `enable-eslint`.

## Documentation

`CONVENTIONS.md` documents the PHP-only configuration (as the counterpart of the existing JS-only section) and the two new frontend inputs.

## Compatibility

All three inputs default to the current behaviour — existing callers are unaffected.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -27,6 +27,11 @@ on:
         required: false
         type: boolean
         default: true
+      enable-npm:
+        description: "Master switch for the npm ecosystem checks (npm legs of security/license). Set false for PHP-only repos without package.json — `npm ci` fails hard without one. Counterpart of enable-php."
+        required: false
+        type: boolean
+        default: true
       enable-psalm:
         description: "Run Psalm static analysis"
         required: false
@@ -67,6 +72,16 @@ on:
         required: false
         type: string
         default: "20"
+      frontend-path:
+        description: "Directory containing package.json + package-lock.json for all npm-side jobs (Vue Quality, Frontend Checks, npm security/license legs). Default: repo root. Use for repos where the frontend lives in a subdirectory (e.g. 'pwa')."
+        required: false
+        type: string
+        default: "."
+      enable-stylelint:
+        description: "Run Stylelint (requires enable-frontend). Set false for repos without a stylelint npm script."
+        required: false
+        type: boolean
+        default: true
       frontend-setup-command:
         description: "Optional ADDITIONAL setup command for the frontend-checks job. The standard root `npm ci` always runs first; this command runs after it and does NOT replace it. Use it to prepare things the root install doesn't cover, e.g. installing a nested package's dependencies: 'cd docusaurus && npm ci' (the nested directory then needs its own package.json + package-lock.json)."
         required: false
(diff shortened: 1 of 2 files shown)


rjzondervan (approved):

rjzondervan:
Not too much files or a too complicated description: this seems fine

WilcoLouwerse (approved):
## Verdict: APPROVE (Standard)

Clean, additive, backward-compatible — new inputs default to current behaviour, skipped legs still report so org-wide required checks stay green, and the `FRONTEND_PATH` env-var pattern keeps caller strings out of interpolation. All 4 CI checks passed.

Three notes to consider — none are blockers, and none map to lines in this PR's diff, so they're here rather than as inline comments.

### 🟡 SBOM job doesn't honor `frontend-path`

The [SBOM job](https://github.com/ConductionNL/.github/blob/6b3aa7816f5f9326ce892bd89028d5c1ae100738/.github/workflows/quality.yml#L2038-L2100) at `.github/workflows/quality.yml:2038-2100` has npm-related steps (`Install npm dependencies` at L2078, `Generate npm SBOM` at L2092, and the SBOM merge at L2097) gated by `enable-frontend` alone. It also doesn't observe `enable-npm`. For a nested-frontend repo like the `woo-website-template-apiv2` example called out in the PR body (`enable-frontend: true`, `frontend-path: pwa`, `enable-sbom: true`), the `npm ci` at L2078 runs at the repo root and would fail — no `package.json` there.

Not a regression (this PR doesn't touch the SBOM job), but the "nested frontend" story is incomplete without it. Either extend `frontend-path` + `enable-npm` to the SBOM job here, or document the limitation in `CONVENTIONS.md` and follow up.

### 🟢 Stale docstring on `frontend-setup-command`

The existing description at [`.github/workflows/quality.yml:86`](https://github.com/ConductionNL/.github/blob/6b3aa7816f5f9326ce892bd89028d5c1ae100738/.github/workflows/quality.yml#L86) says _"The standard root `npm ci` always runs first"_. After this PR, that `npm ci` runs in `frontend-path`, not always at the repo root. Worth a wording tweak so nested-frontend callers know their `frontend-setup-command` executes relative to `frontend-path` (and their examples like `cd docusaurus && npm ci` resolve from there).

### 🟢 Two switches for PHP-only repos

`CONVENTIONS.md` correctly tells PHP-only repos to set BOTH `enable-frontend: false` AND `enable-npm: false`. Fine as designed, but a computed default — `enable-npm` defaulting to `enable-frontend` — would remove the coordination requirement. Design preference; leave as-is if you'd rather keep the switches orthogonal.

---

Nice work — the env-var pattern for `FRONTEND_PATH`, the exit-0-inside-matrix pattern for keeping required checks satisfied, and the CONVENTIONS.md updates make this easy to review.

Issue #61, opened 2026-05-13
Allow CI bot to push auto-generated files (.coverage-baseline, SBOM) to ruleset-protected branches


rubenvdlinde:
## Problem

The shared `quality.yml` workflow's **`update-baseline`** job (and, if/when we start committing SBOMs back, a similar SBOM-commit step) runs on push to `main`/`development`, regenerates `.coverage-baseline`, commits it as `github-actions[bot]`, and `git push`es. On any repo whose `development`/`main` branch is covered by the org rulesets (1–2 required reviews), that push is rejected:

```
remote: error: GH013: Repository rule violations found for refs/heads/development.
error: failed to push some refs to 'https://github.com/ConductionNL/<app>'
```

So the job fails ⇒ the whole "Code Quality" workflow goes red, even though every actual quality gate passed. Currently visible on `ConductionNL/procest` (which has `enable-coverage-guard: true`); it will hit any app that opts into the coverage guard.

## Asks

1. Let the CI bot push the auto-generated convenience files (`/.coverage-baseline`, and the SBOM file path if we ever start committing it) to `development`/`main` without a review, on every app repo.
2. **(Interim, already done in `quality.yml` — ConductionNL/.github#62):** the `git push` in the `update-baseline` job is now non-blocking — it logs a `::warning::` instead of failing the build. Once (1) is in place that can stay (the warning just won't fire) or be tightened back to a hard failure.

## Where / how to configure this

The rulesets are **org-level**, not per-repo:

- `https://github.com/organizations/ConductionNL/settings/rules` → Org Settings → *Code, planning, and automation* → **Repository → Rulesets** → open the `dev` ruleset (and `beta`, `main`). Each has a **"Bypass list"** section ("Add bypass").

Wrinkles to be aware of:

- **`github-actions[bot]` can't be added as a bypass actor directly.** Ruleset bypass actors are: repo roles (Admin / Maintain / Write), Teams, Deploy keys, or **GitHub Apps**. The `${{ github.token }}` the workflow uses *is* `github-actions[bot]`, and it isn't selectable.
- **Bypass is all-or-nothing per actor** — you can't scope a bypass actor to "only push `/.coverage-baseline`". In practice that job only ever pushes that one file, so it's acceptable; but if path-scoping matters, the cleaner option is to have the job open a PR (see option B below).

### Option A — dedicated GitHub App as a ruleset bypass actor (recommended)

[...]

PR #128, opened 2026-08-03
ci(quality): features.json generated at commit time — CI verifies and blocks, never commits

remko48:
## What

The quality pipeline no longer commits to the branches it checks. `docs/features.json`
is now generated **at commit time** by a committed git hook; CI only verifies — and
verification failures are merge blockers.

- **`features-extract`** (push events) rewritten: read-only. Regenerates in memory,
  **hard-fails** when `docs/features.json` is stale, and attaches the regenerated file
  as a run artifact. No more `git commit`/`git push` from CI.
- **`features-check`** (PR events) unchanged — already a read-only `--check` gate.
- Both jobs feed the `Quality Report` gate (`needs:` + fail step, already on main), so a
  stale features.json fails `quality / Quality Report` and **cannot be merged** —
  hook or no hook.
- Comments and input descriptions updated to match; CONVENTIONS.md gains the full
  per-repo rollout guide (verbatim hook, activation via npm `prepare` /
  composer `post-install-cmd`, husky + manifest-less variants, enforcement semantics).

## Why

The auto-commit was well-intentioned (devs forgot to regenerate) but self-defeating:

1. The bot commit carried `[skip ci]` and became the PR head → **zero workflows ran
   for the head SHA**, the PR lost all its checks (observed on nextcloud-vue#569).
2. Every bot push **dismissed reviewer approvals** (org ruleset dismisses stale
   reviews on push).
3. On protected branches the push bounced off branch protection anyway (#61 — now
   obsolete, can be closed).

A quality pipeline must never mutate the branch it is judging. Generation belongs in
the developer's commit, before the checks run; CI enforces.

## SBOM

Already correct and needs no per-repo setup: generated in CI, hard-failing validation
(Grype CVE scan + composer/npm audit), published as run artifact + release asset, and
**intentionally never committed** (SBOMs embed timestamps/serial numbers — a committed
copy would be perpetually stale). Failures block through the same Quality Report gate.
Input description updated accordingly (it still claimed "commit to repo").

## Rollout (per spec-bearing repo, we do this ourselves)

1. Copy `.githooks/pre-commit` from CONVENTIONS.md (reference implementation: openregister).
2. Wire activation into every manifest the repo has (`package.json` "prepare" /
   composer `post-install-cmd`); husky repos embed the snippet in `.husky/pre-commit`.
3. `git config core.hooksPath .githooks` once per existing clone.
4. Regenerate once so the repo enters the enforced state green.

Repos without `openspec/specs` keep `enable-features-extract: false` — nothing to do.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -173,12 +173,12 @@ on:
         type: boolean
         default: false
       enable-sbom:
-        description: "Generate CycloneDX SBOM, validate (CVE + license), and commit to repo"
+        description: "Generate CycloneDX SBOM and validate it (Grype CVE scan + composer/npm audit — hard failures). Published as a run artifact and release asset; intentionally never committed to the repo."
         required: false
         type: boolean
         default: true
       enable-features-extract:
-        description: "Scan openspec/specs/ and (re)generate docs/features.json — auto-commits on feature branches, warns on protected branches"
+        description: "Verify docs/features.json is in sync with openspec/specs/ — READ-ONLY, hard-fails when stale (features-check on PRs, features-extract on pushes; both gate the Quality Report). Generation happens at commit time via the repo's .githooks/pre-commit — see CONVENTIONS.md."
         required: false
         type: boolean
         default: true
@@ -2503,7 +2503,7 @@ jobs:
           files: sbom.cdx.json
 
   # ╔══════════════════════════════════════════════╗
-  # ║  STAGE 5 — Features extract & commit          ║
+  # ║  STAGE 5 — Features verification (read-only)  ║
   # ║                                               ║
   # ║  Scans openspec/specs/*/spec.md for           ║
   # ║  capabilities whose YAML frontmatter declares  ║
@@ -2512,10 +2512,12 @@ jobs:
   # ║  Features & Roadmap surface consumes via       ║
   # ║  initial-state injection.                      ║
   # ║                                               ║
-  # ║  Branch policy mirrors update-baseline:        ║
-  # ║  feature branches → auto-commit + push;        ║
-  # ║  protected branches → warn only (see           ║
-  # ║  ConductionNL/.github#61).                     ║
+  # ║  CI never commits: generation happens at       ║
+  # ║  COMMIT TIME via each repo's committed          ║
+  # ║  .githooks/pre-commit (CONVENTIONS.md). Both    ║
+  # ║  jobs below only verify and HARD-FAIL when     ║
+  # ║  stale — via the Quality Report gate that      ║
+  # ║  blocks the merge.                             ║
   # ╚══════════════════════════════════════════════╝
 
   # On pull_request events, run the extractor in `--check` mode only — fail
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -2550,18 +2550,24 @@ jobs:
       - name: Verify docs/features.json is up to date
         run: python .conduction-shared/scripts/extract-features.py --app-root . --check
 
+  # READ-ONLY by design. This job used to auto-commit + push the regenerated
+  # docs/features.json back to the branch — a quality pipeline mutating the
+  # branch it is judging. That moved the PR head from under the checks
+  # (with [skip ci] it stripped ALL checks from the PR), dismissed reviewer
+  # approvals via the org ruleset's dismiss-stale-on-push, and on protected
+  # branches the bot push bounced anyway (#61). Generation now happens at
+  # COMMIT TIME in each repo via the committed .githooks/pre-commit hook
+  # (see CONVENTIONS.md § features.json is generated at commit time); CI
+  # only verifies and hands back the regenerated file as an artifact.
   features-extract:
     if: ${{ inputs.enable-features-extract && !cancelled() && github.event_name != 'pull_request' }}
     runs-on: ubuntu-latest
     name: "Features Extract"
     permissions:
-      contents: write
+      contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.head_ref || github.ref_name }}
-          token: ${{ github.token }}
 
       - name: Setup Python
         uses: actions/setup-python@v5
@@ -2581,16 +2587,19 @@ jobs:
       - name: Regenerate docs/features.json
         run: python .conduction-shared/scripts/extract-features.py --app-root .
 
-      - name: Commit if changed
+      - name: Fail if stale
         run: |
-          git config user.name "github-actions[bot]"
-          git config user.email "github-actions[bot]@users.noreply.github.com"
           if git diff --quiet docs/features.json 2>/dev/null && ! git status --porcelain docs/features.json | grep -q '^??'; then
-            echo "docs/features.json unchanged, nothing to commit."
+            echo "docs/features.json is up to date."
           else
-            git add docs/features.json
-            git commit -m "ci: regenerate docs/features.json from openspec/specs/ [skip ci]"
-            # Non-blocking on protected branches: branch-protection rulesets reject
-            # the bot push on main/beta/development. Tracked in ConductionNL/.github#61.
-            git push || echo "::warning::Could not push the regenerated docs/features.json — branch protection rejected the bot push (see ConductionNL/.github#61). Run scripts/extract-features.py locally and commit manually."
+            echo "::error::docs/features.json is out of date with openspec/specs/. Regenerate it locally (the committed .githooks/pre-commit hook does this automatically — see CONVENTIONS.md) and commit it. The regenerated file is attached as the 'features-json' artifact of this run."
+            exit 1
           fi
+
+      - name: Upload regenerated file
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: features-json
+          path: docs/features.json
+          retention-days: 7
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -2590,6 +2590,7 @@ jobs:
         run: python .conduction-shared/scripts/extract-features.py --app-root .
 
       - name: Fail if stale
+        id: fail-if-stale
         run: |
           if git diff --quiet docs/features.json 2>/dev/null && ! git status --porcelain docs/features.json | grep -q '^??'; then
             echo "docs/features.json is up to date."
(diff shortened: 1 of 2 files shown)


rjzondervan (approved):

.github/workflows/quality.yml
```diff
@@ -2581,16 +2589,19 @@ jobs:
...
-            git add docs/features.json
-            git commit -m "ci: regenerate docs/features.json from openspec/specs/ [skip ci]"
-            # Non-blocking on protected branches: branch-protection rulesets reject
-            # the bot push on main/beta/development. Tracked in ConductionNL/.github#61.
-            git push || echo "::warning::Could not push the regenerated docs/features.json — branch protection rejected the bot push (see ConductionNL/.github#61). Run scripts/extract-features.py locally and commit manually."
+            echo "::error::docs/features.json is out of date with openspec/specs/. Regenerate it locally (the committed .githooks/pre-commit hook does this automatically — see CONVENTIONS.md) and commit it. The regenerated file is attached as the 'features-json' artifact of this run."
+            exit 1
           fi
+
+      - name: Upload regenerated file
```
WilcoLouwerse, line 2602: 🟡 **Concern — `if: failure()` is too broad; can attach a stale file when the regenerate step itself fails.**

`if: failure()` fires when *any* prior step failed, including `Regenerate docs/features.json`. If regenerate itself crashes (malformed `spec.md`, `pyyaml` import error, etc.), the `Fail if stale` step is skipped, but this upload still runs — attaching whatever `docs/features.json` sits on disk. That's likely the committed stale version (no fresh write happened), or nothing (if the file was never committed). It gets uploaded under the label `features-json` as if it were the fresh regeneration.

The error message on the staleness step tells the developer to "download the artifact and commit it." Following that advice after a regenerate-step failure would commit the same stale content the CI just flagged.

**Suggested fix — narrow the trigger:**

```yaml
      - name: Fail if stale
        id: fail-if-stale
        run: |
          ...

      - name: Upload regenerated file
        if: steps.fail-if-stale.outcome == 'failure'
        uses: actions/upload-artifact@v4
        with:
          name: features-json
          path: docs/features.json
          retention-days: 7
```

Now the upload only fires when the staleness check specifically failed — which is exactly the case where the on-disk file *is* the fresh regeneration.

CONVENTIONS.md
```diff
@@ -133,6 +133,73 @@ A release or critical merge is blocked **right now**, the real fix (aligning the
...
 A `.npmrc` containing `legacy-peer-deps=true` without a dated explanation and a linked issue should be treated as a bug and flagged in review.
 
+#### features.json is generated at commit time, never by CI
+
+`docs/features.json` (the commercial capability list derived from `openspec/specs/`) is regenerated **on the developer's machine at commit time**, by a committed pre-commit hook. CI only verifies:
+
+- `features-check` (PRs) and `features-extract` (pushes) are both **read-only gates** — they regenerate in memory, fail if the committed file is stale, and attach the regenerated file as a run artifact. Neither ever commits or pushes.
+- The pipeline previously auto-committed the regenerated file back to the branch. That is forbidden now and must not come back: a CI push moves the PR head out from under its checks (with `[skip ci]` it stripped **all** checks from the PR), dismisses reviewer approvals via the org ruleset's dismiss-stale-on-push, and bounced off branch protection on protected branches anyway (#61). A quality pipeline must never mutate the branch it is judging.
+
+Repo setup (reference implementation: `openregister`):
```
WilcoLouwerse: 🟡 **Concern — the referenced implementation (`openregister`) doesn't ship the hook.**

Verified via the GitHub API: `openregister/.githooks/pre-commit` returns HTTP 404 on `main` and doesn't appear in any branch listing. A tree scan for `pre-commit` under `openregister` also returns nothing. So readers following "reference implementation: `openregister`" have nothing to cross-reference against — they land on the verbatim block on L159 instead.

Two ways out:

1. **Ship the hook to `openregister`** in the same batch (or a companion PR merged first) so the phrase describes reality.
2. **Drop "reference implementation: `openregister`"** here and either name a repo that *does* have it, or route readers straight to the verbatim block on L159 as the canonical source.

Current state creates the exact drift the doc is trying to prevent ("paper says X, code doesn't").

CONVENTIONS.md
```diff
@@ -133,6 +133,73 @@ A release or critical merge is blocked **right now**, the real fix (aligning the
...
+```sh
+#!/bin/sh
+# Regenerates docs/features.json whenever staged changes touch openspec/specs/
+# or the features overlay. Best-effort: warns but never blocks the commit —
+# the CI gate (features-check/features-extract → Quality Report) enforces.
+
+if git diff --cached --name-only | grep -qE "^openspec/(specs/|features\.overlay\.json)"; then
+  CACHE=".git/extract-features.py"
+  curl -sf --max-time 10 \
+    https://raw.githubusercontent.com/ConductionNL/.github/main/scripts/extract-features.py \
```
WilcoLouwerse, line 170: 🟢 **Minor — supply-chain surface: hook fetches from `main` at commit time.**

Every commit that touches `openspec/specs/` triggers a fresh download of `extract-features.py` from `main`, which then executes as the developer. Any actor who lands code in `ConductionNL/.github/main` gets execution on every consuming developer's machine on their next openspec-touching commit.

This mirrors the reusable workflow's own `ref: main` checkout of `.conduction-shared` (quality.yml L2582), so the trust model is *consistent*. But the blast radius differs — an ephemeral runner vs. developer laptops with SSH keys, cached credentials, and access tokens.

Options for hardening (any one is enough):
- Pin the fetch to a tagged release (`refs/tags/quality-v1`) and bump on script changes.
- Verify a committed SHA-256 checksum after fetching (fail closed if mismatch).
- Treat the cached `$CACHE` as the primary and refresh only on explicit version bump.

At minimum, document this trust boundary in CONVENTIONS.md so downstream repos adopting the pattern know what they're inheriting. Not blocking on its own since it matches existing CI practice — flagged for future hardening.

CONVENTIONS.md
```diff
@@ -133,6 +133,73 @@ A release or critical merge is blocked **right now**, the real fix (aligning the
...
+    -o "$CACHE" 2>/dev/null || true
+
+  if [ -f "$CACHE" ]; then
+    if command -v python3 >/dev/null 2>&1; then PY="python3";
+    elif command -v py >/dev/null 2>&1; then PY="py -3";
+    else PY="python"; fi
+
+    if $PY "$CACHE" --app-root . >/dev/null 2>&1; then
```
WilcoLouwerse: 🟢 **Minor — silencing stderr masks the actual failure reason.**

`$PY "$CACHE" --app-root . >/dev/null 2>&1` redirects both stdout AND stderr to `/dev/null`. When the extractor legitimately errors (invalid YAML frontmatter in a `spec.md`, malformed capability entry, schema violation), the developer only sees the fallback warning:

> `pre-commit: WARNING — could not regenerate docs/features.json (python or pyyaml missing?). CI features-check will verify.`

…which mis-attributes the cause: dependencies are fine, the spec file has a problem. The developer then wastes time chasing a nonexistent pyyaml issue instead of fixing the actual spec.

**Suggested tweak** — keep stdout quiet but let stderr surface:

```sh
if $PY "$CACHE" --app-root . >/dev/null; then
```

Same happy-path behavior (silent success), but real errors reach the developer's terminal on failure.

WilcoLouwerse (changes requested):
## Verdict: REQUEST_CHANGES (Strict) — 2 🟡 + 2 🟢

Strong direction — this removes a real anti-pattern (CI mutating the branch it judges), tightens permissions to `contents: read`, and adds a recovery artifact. Requesting changes on two hygiene items that ride along:

- 🟡 **[Missing reference implementation](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042799)** (`CONVENTIONS.md` L143) — cites `openregister` as reference but `openregister/.githooks/pre-commit` returns 404 on `main`. Ship the hook there or drop the phrase and point at the verbatim block on L159.
- 🟡 **[Over-broad `if: failure()`](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042791)** (`quality.yml` L2601) — uploads whatever's on disk when *any* prior step fails, so a `Regenerate` crash attaches the stale committed file under the `features-json` label. Narrow to `steps.fail-if-stale.outcome == 'failure'`.

Plus 2 🟢 minors on the hook — non-blocking but worth folding in:
- 🟢 **[Supply-chain surface — fetches `main` at commit time](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042810)** (`CONVENTIONS.md` L170)
- 🟢 **[Silenced stderr masks the actual failure reason](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042820)** (`CONVENTIONS.md` L178)

Verified in this pass:
- `Quality Report` `needs:` array (L2073) includes both `features-check` and `features-extract` ✓
- `scripts/extract-features.py` exists at the expected path on `.github` `main` ✓
- Job-scoped `permissions: contents: read` correctly narrows the token below any workflow-level defaults ✓
- All 4 CodeQL checks green ✓

.github/workflows/quality.yml
WilcoLouwerse: Nicely fixed in 3c5c7111 — and the `failure() && steps.fail-if-stale.outcome == 'failure'` combination is more correct than my original suggestion. Without the `failure() &&` prefix, GitHub silently implies `success() && …` and the upload step would never run after a preceding step failure. The inline comment above the `if:` documents exactly that trap. Thanks.

CONVENTIONS.md
WilcoLouwerse: Addressed in 3c5c7111 — the phrase `reference implementation: openregister` was swapped for "the verbatim hook below is the canonical source — copy it, don't reinvent it", which side-steps the drift risk entirely. Thanks.

CONVENTIONS.md
WilcoLouwerse: Addressed in 3c5c7111 — new **Trust boundary** paragraph documents exactly the runner-vs-dev-machine blast-radius delta and names both hardening paths (tag pin / SHA-256 checksum verify) as the escalation route when it's needed. Explicit trust-boundary docs are the minimum I was after. Thanks.

CONVENTIONS.md
WilcoLouwerse: Addressed in 3c5c7111 — `2>&1` dropped, and the fallback warning updated to "see error above; missing python/pyyaml also lands here", which correctly re-frames the message as "an error just printed; here's what to check" rather than mis-attributing every failure to a missing dependency. Thanks.

WilcoLouwerse (approved):
## Verdict: APPROVE (Quick re-review) — all 4 findings resolved in `3c5c7111`

Fix commit lands all four findings cleanly:

- 🟡 → ✅ **[Over-broad `if: failure()`](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042791)** — `id: fail-if-stale` added, gate narrowed to `failure() && steps.fail-if-stale.outcome == 'failure'`. The `failure() &&` prefix was necessary (my original suggestion was incomplete — without a status function, GitHub silently implies `success() &&` and the step would have been skipped); the inline comment above the `if:` documents exactly that trap.
- 🟡 → ✅ **[Missing reference implementation](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042799)** — "reference implementation: openregister" swapped for "the verbatim hook below is the canonical source — copy it, don't reinvent it". Drift risk gone.
- 🟢 → ✅ **[Supply-chain surface](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042810)** — new **Trust boundary** paragraph documents the runner-vs-dev-machine blast-radius delta and names both hardening paths (tag pin / SHA-256 verify) as the escalation route.
- 🟢 → ✅ **[Silenced stderr](https://github.com/ConductionNL/.github/pull/128#discussion_r3704042820)** — `2>&1` dropped, warning message updated to point at the real error above.

CI still green on `3c5c7111` (4/4 CodeQL). Zero regressions. Threads resolved.

Issue #160, opened 2026-08-05
gate-7 cannot see a tenancy guard that refuses with 404; gate-9 rejects token-auth PublicPage; shared /tmp log paths make verdicts non-deterministic


rubenvdlinde:
Re-derived openregister's gates against the **full tree** (the enablement PR was diff-scoped over three CI files, so 58 gates "passed" having read no `lib/`). 19 gates fail. Gates 6/7/8/9 produced 27 findings; I read the call path for every one. **Four were real and are fixed in ConductionNL/openregister#TBD.** The rest split into three gate defects, below.

---

## 1. gate-7 cannot see an organisation-scoping guard that refuses with "not found"

**This is the important one, and it is not closed by #149.** Measured on `.github@main` (which has #149) — openregister's gate-7 count is identical at v1.0.1 and at main: **18 findings, and 17 of them are false positives.**

`#149`'s Pattern 4 propagates only through `_STRICT_GUARD_BODY_RE`: an explicit 401/403, `OCSForbiddenException`, `isAdmin`, an `authorize*`/`require*`/`ensure*`/`assert*`/`guard*` call, or an anonymous-session rejection. A bare `throw` and a 404 deliberately do not seed a chain, on the reasoning that "`NotFoundException` is not an authorisation guard".

For a **multi-tenant** codebase that reasoning inverts. OpenRegister's `FlowService` is explicit about it:

```php
/**
 * A flow the caller may not see raises the SAME exception as a flow that
 * does not exist. Distinguishing them would turn every read into an oracle
 * for enumerating other tenants' flow ids.
 */
public function find(string $uuid): Flow
{
    $flow = $this->mapper->findByUuid($uuid);
    if ($flow->belongsTo($this->activeOrganisation()) === false) {
        throw new DoesNotExistException('No such flow');
    }
    return $flow;
}
```

That is the guard. It is *deliberately* indistinguishable from a 404, because a 403 here leaks the existence of another tenant's object. Gate-7 therefore flags exactly the code that got tenancy right, and would be quieted by making it leak.

Same for the list side: `FlowService::findAll()` returns `[]` when no organisation resolves. Silent narrowing, no deny signal, invisible to the gate.

[...]

Issue #159, opened 2026-08-05
The whole fleet pins hydra-gates-ref: v1.0.1, which predates #147/#148/#149 — 16 dead gates are still live in CI everywhere


rubenvdlinde:
All 22 repos that set `enable-hydra-gates: true` also pin:

```yaml
hydra-gates-ref: v1.0.1
```

`v1.0.1` is `f4d9756` (2026-08-03). Everything since is invisible to CI:

| commit | tag | in `v1.0.1`? |
|---|---|---|
| `f4d9756` fix(gate-4): audit the LOCK (#135) | **v1.0.1** | ✅ |
| `d9645e3` gate-53 ignored the diff scope; gate-22 printed `FAIL — 0`; gate-33 had never run (#139) | v1.1.0 | ❌ |
| `549ad1a` **16 gates reported PASS when their helper never ran** (#147) | — | ❌ |
| `0d3e589` produce `tests/axe/report.json` so gate-33 can actually run (#148) | — | ❌ |
| `fc70ea6` gate-7 follows delegation; gates 6/7 stop passing on an empty scope (#149) | — | ❌ |

So every in-CI Hydra Gates run on the fleet today is executing a script in which **16 gates report PASS when their helper never ran**, and in which gates 6 and 7 report PASS on an empty scope. That is the dead-gate shape those PRs were written to close, still live everywhere.

Confirmed for: app-versions, decidesk, docudesk, doriath, hermiq, larpingapp, launchpad, nextcloud-app-template, nldesign, openbuild, opencatalogi, openconnector, openregister, petstore, pipelinq, planix, portaliq, procest, scholiq, shillinq, softwarecatalog, zaakafhandelapp.

```
$ for r in …; do gh api repos/ConductionNL/$r/contents/.github/workflows/code-quality.yml?ref=development \
    -q .content | base64 -d | grep hydra-gates-ref; done
hydra-gates-ref: v1.0.1     # ×22, no exceptions
```

**Two things needed:**

1. **Cut a tag containing #147/#148/#149** (`v1.2.0`) and bump the fleet to it. `v1.1.0` is not enough — it predates all three.
2. **Decide whether the pin should be a moving minor** (`v1`) rather than an exact patch. A patch pin means every gate fix needs a 22-repo bump PR wave, which is exactly why the fleet is 4 fixes behind after two days.

Related: the full-tree numbers I am reporting elsewhere use `main`, so they are **not** comparable to what CI prints today — worth stating explicitly wherever both appear.

[...]

PR #175, opened 2026-08-05
fix(gates): 46/40/28/9/7 were majority-false, and the runner corrupted its own measurements

### LINKED ISSUE #158: hydra-gates: five false-positive modes (~130 of 276 fleet findings), one with dangerous remediation advice — plus the runner is unsafe to run concurrently
Re-running the gates against the **full tree** across 21 fleet repos (excluding openregister) produced **276 failing gates**. Triaging the authorisation gates line by line — reading every call path — turned up **five distinct false-positive modes**. Together they account for **~130 of the 276**, and one of them ships remediation advice that would *introduce* a vulnerability if followed.

Measurement: gate script `ConductionNL/.github@ad00986` (`main`, includes #147/#148/#149), `ajv` installed under `hydra-gates/scripts/lib`, each repo at its current `origin/development`, **no** `--scope-to-diff`. Each run was executed in its own mount namespace with a private `tmpfs /tmp`, because the runner writes its detail logs to hardcoded `/tmp/hydra-gate-<name>.log` paths and reads the counts back out of them — two concurrent runs silently corrupt each other's numbers (see item 6).

---

## 1. gate-38 `skip-link` — fails in **21 of 21** repos

Uniformity across independent inputs points at the instrument, and it is.

The gate greps a root component for a literal `<NcContent>` / `<NcAppContent>`. The fleet has migrated to `<CnAppRoot>` (nc-vue), and **`CnAppRoot` renders `<NcContent>` internally** — `src/components/CnAppRoot/CnAppRoot.vue:60`. Every manifest-driven app root therefore fails a check it satisfies.

Second cause in the same gate: it scopes `templates/settings/*.php`. Those are Nextcloud settings-framework templates; the framework owns the page shell, so such a file can *never* contain a skip-link and can never pass.

**Fix:** accept `<CnAppRoot>` (and any component that transitively renders `NcContent`) as a shell; drop `templates/settings/*.php` from scope.

## 2. gate-9 `semantic-auth`, rule `public-page-annotation-with-auth-body` — **36 of 39** findings, and the advice is dangerous

The gate tells you: *"remove `#[PublicPage]` or remove body auth check"*. Every one of the 36 is a **webhook or public-portal endpoint that correctly bypasses NC session auth and authenticates the caller itself**. Following either half of that advice breaks the endpoint or removes its only authentication.

Verified call paths:

- `openconnector PaymentsController::webhook` — HMAC signature verify with constant-time compare + timestamp tolerance, 401 before any state change. Its docblock already says *"the signature check IS the auth body for this route"*.
- `portaliq ContributionController::inbox` (and 9 siblings) — resolves a portal `subject()`, 401 when absent; per-row subject + tenant + trust-boundary filtering.
- Same shape: `openconnector` DSO / IwmoIjw / Notificaties / NotifyNl / OpenFormulieren / Peppol / StufZkn inbound, `procest CaseFederationController` ×3, `hermiq EgressAuthorize` + `McpRun`, `launchpad PublicShareController` ×2, `doriath ApplicationController::create` + `ApplicationTokenController::exchange`, `scholiq PaymentTransactionController::callback`, `shillinq PortalPaymentInitiationController::initiate`, `softwarecatalog` ×2, `petstore` ×1.

**Fix:** the rule must distinguish *"public but self-authenticating"* (signature verify, bearer/portal token, capability token → **correct**) from *"`#[PublicPage]` while relying on an `IUserSession` session check"* (→ the real mismatch). Only 3 of the 39 findings are the `no-admin-required-annotation-with-admin-body` shape the gate was actually built for, and those were real — fixed in openbuild#127.

## 3. gate-7 `no-admin-idor` — **14 of 15** findings false, even after #149

#149 taught the gate to follow delegation; two blind spots remain.

**(a) A guard that returns `bool`/`null` instead of throwing or returning 401/403.** `hermiq` ×7: `AgentVersionController::index`/`diff` → `loadAccessibleAgent($id, $userId)` → `canUserAccessAgent()` (private/owner/invited); `AgentsController::index` applies `canUserAccessAgent()` **per row inside the result loop**. `_GUARD_HELPER_NAME_RE` requires the name to end in `Admin|Access|Permission|Permitted|Owner|Allowed|Authorised` — `canUserAccessAgent` ends in `Agent`, `loadAccessibleAgent` starts with `load`; and `_HELPER_GUARD_BODY_RE` wants a throw or a 401/403/404, which a `bool`-returning predicate never has.

**(b) Endpoints with no object identifier at all.** `nldesign CatalogController::tokenSets` (public catalogue, no args), `nldesign ContrastController::evaluate` (pure colour math on caller-supplied input), `procest AssistantController::availability` (boolean feature probe), `hermiq AgentsController::create`/`stats`/`tools`. There is no object to be *insecure-direct-referenced*; the gate should not ask for a per-object guard where no object is addressed.

`doriath CxpRelayController::put`/`get` are a capability-URL relay (the 24-char `ISecureRandom` pairing id is the bearer credential, payloads are sealed) and `SettingsController::getPolicy` is a documented read-open-to-all-authenticated. Both false.

**The one real finding** was `docudesk AnonymizationController::updateRelation` — genuinely unguarded through four service hops to an unscoped primary-key write. Fixed in docudesk#382. So gate-7's true-positive rate here is **1 in 15** — high enough to be worth keeping, low enough that the noise will train people to skip it.

## 4. gate-14 `route-reachability` — **22 of 30** findings are AppHost delegation

`openbuild` (7), `shillinq` (7), `doriath` (5), `planix` (3) are reported as having unrouted `dashboard#page`, `dashboard#catchAll`, `settings#index/create/load`, `preferences#getPreference/setPreference`. All are supplied by `OCA\OpenRegister\AppHost\Routes::standard()`. The gate parses only the literal array in `appinfo/routes.php`, so it cannot see routes contributed by the shared builder.

Tell: `docudesk` *passed* this gate only because it happens to carry a literal `$canonicalRoutes` fallback array for the openregister-absent case. Passing should not depend on that.

**Fix:** when `appinfo/routes.php` calls `Routes::standard()`, fold in `Routes::canonicalRoutes()` before deciding.

The other 8 findings are **real** and worth keeping — see item 5.

## 5. gate-6 `orphan-auth` — 2 of 5 false

- `launchpad LiveTileController::validateSource` — a **routed** controller method (`liveTile#validateSource`, POST, `appinfo/routes.php:557`). Its caller is the NC router, not PHP code.
- `pipelinq LogBerichtenboxAdapter::checkMailbox` — implements `BerichtenboxAdapterInterface::checkMailbox`; callers go through the interface.

**Fix:** exempt routed controller methods and declared-interface implementations.

The other three are **real** decidesk#60-shaped orphans — implemented, unit-tested, never wired into any production path — and are being filed against their repos: `procest BewijsstukService::assertMutable`, `pipelinq PortalTenantService::isSelfSignupAllowed`, `pipelinq ZgwCoexistenceValidator::validateWritePath`.

## 6. The runner is not safe to run twice at once

`run-hydra-gates.sh` writes ~50 detail logs to hardcoded `/tmp/hydra-gate-<name>.log` and derives each verdict from `wc -l` on that shared file. Two concurrent runs — trivially easy across repos, or a stray backgrounded run nobody noticed — produce numbers that are *plausible but wrong*, in both directions.

This bit me during this exercise: a first measurement pass reported `softwarecatalog` with 7 gate-7 IDOR findings. Running `check_no_admin_idor.py` directly against the same tree returned **zero** — the 7 belonged to `hermiq`, running in parallel. `procest` moved 22 → 21, `softwarecatalog` 18 → 16, `zaakafhandelapp` 8 → 9 once the runs were isolated. Only `gate-6` was immune, because it is the one gate already using `mktemp`.

**Fix:** derive a per-run log directory once (`mktemp -d`, or honour `TMPDIR`) and hang every gate's log off it — the pattern gate-6 already uses. Cheap, and it removes a whole class of "measured, but wrong" from every future fleet sweep.

---

### Suggested priority

1. **item 6** — until it lands, no fleet-wide number from this runner is trustworthy.
2. **item 2** — the remediation text is actively harmful.
3. **items 1, 4** — pure noise, ~43 findings, and they are what makes people stop reading gate output.
4. **items 3, 5** — narrow the guard vocabulary and the scope.

Nothing was suppressed, baselined or excluded anywhere in this exercise.

### LINKED ISSUE #171: gate-28 (license-triangle) reads only the FIRST @license tag and ignores SPDX-License-Identifier — 174 files carried an AGPL claim behind a green gate
A fleet-wide EUPL-1.2 sweep measured **19,200 licence declarations, 829 of them non-EUPL**. Fixing them exposed two defects in gate-28 itself.

## 1. The gate is blind to a second licence in the same file

**174 files declare BOTH `EUPL-1.2` and an AGPL licence in the same file** — launchpad 123, openregister 47, openconnector 4.

gate-28 reads only the **first** `@license` tag and **ignores `SPDX-License-Identifier:` entirely**. So a file whose PHPDoc says EUPL-1.2 and whose SPDX line says `AGPL-3.0-or-later` passes.

**A green gate-28 was never evidence a repo was clean.** Most of those 174 files were passing.

Suggested fix: collect *every* licence declaration in a file — `@license` tags and `SPDX-License-Identifier:` lines alike — and fail when they disagree with each other or with the repo's declared licence. Disagreement *within* one file is the stronger signal, because it means someone edited one and not the other.

Note there is a legitimate exception to encode: **assertion strings in tests**. `nldesign`'s `MarianneFontTest.php` and `ClaimAccuracyTest.php` contain non-EUPL identifiers as test data, not declarations — changing them breaks the tests.

## 2. A single NUL byte turns a real PASS into a false RED

`shillinq/lib/Service/InventoryValuationReportService.php` contains a raw NUL at line 515, inside a PHPDoc describing a `sku\0warehouse` composite key. Consequences:

- `file` reports the PHP source as `data`
- `grep` switches to binary mode and prints `Binary file … matches` instead of the line
- **gate-28 then reads the FILE PATH as the licence value** and reports a violation

The file's header was already correct. This is a **false RED manufactured by the gate's own text handling**.

The same raw-NUL idiom is in live code elsewhere and will hit this again:
- `doriath/src/import/model.js:117`
- `openbuild/src/services/manifestValidation/documentAttachments.js:136`

Suggested fix: pass `-a`/`--text` to grep (or read with an explicit encoding) so a NUL cannot change the parse, and reject a "licence value" that is a path.

## 3. Unrelated but found in the same pass: GPLv3 inside an EUPL-1.2 repo

`openregister/phpmetrics-deps/` is **39 tracked files of generated report output that `.gitignore` already lists** — committed anyway. Among them:

- `js/clusterize.min.js` — **GPLv3**, © 2015 Denis Lukov
- `js/sort-table.min.js` — MIT, © 2006-2013 Tyler Uebele
- `css/milligram.min.css` — MIT, © 2016 CJ Patoilo

GPLv3 tracked inside an EUPL-1.2 repository is a compatibility question, not a lint finding. The clean fix is `git rm -r --cached phpmetrics-deps/` since it is already gitignored and is build output.

Refs the fleet sweep; gate-28 currently passes in all 28 app repos, which is precisely why defects 1 and 2 are worth fixing before that green is trusted.

rubenvdlinde:
## Why

**1,726 real findings across the fleet cannot be burned down until these gates are honest.** The gates are diff- AND file-scoped, so touching a file drags in all of its pre-existing findings — and when the residue is entirely false positives there is no honest way to green the PR. Two burn-down PRs (opencatalogi#808, docudesk#385) are open and stuck in exactly that state.

The sharpest example: clearing gate-40's residue would mean adding `aria-label` to `NcCheckboxRadioSwitch` elements that **already name themselves from their default slot**. `aria-label` overrides the visible label. The only way to close an accessibility gate was to ship an accessibility regression.

## Measured, 21 fleet repos at `origin/development`, baseline = v1.4.0 (`46cc1c8`)

| gate | before | after | Δ |
|---|---:|---:|---:|
| 46 spec-anchor-existence | 1,995 | **918** | −54% |
| 40 form-label-association | 1,211 | **517** | −57% |
| 9 semantic-auth | 45 | **11** | −76% |
| 7 no-admin-idor | 32 | **26** | −19% (partial) |
| 28 license-triangle | 0 | 0 | preventive — see below |

Note these are **not** the 2,630 / 1,365 in #158: those predate `gh_slugify` (#165), which is in v1.4.0 and in no earlier tag. Measured honestly against what `main` actually does today.

## Never make a gate pass by making it blind

Every relaxation ships **paired** with the true-positive case it must not swallow. Four new suites (46, 40, 28, 9) plus 8 new gate-7 cases, all auto-discovered by `tests/run-helper-suites.sh`.

Each relaxed predicate was **mutation-checked in both directions**:

| mutation | result |
|---|---|
| gate-46 `has_anchor` → always True | 18 failures |
| gate-46 digit guard removed | 2 failures |
| gate-40 label check → always True | 10 failures |
| gate-40 slot check → always True | 1 failure |
| gate-40 `:id`/`:for` → "is bound at all" | 2 failures |
| gate-7 tenancy guard → always True | **8** failures |
| gate-7 tenancy guard → always False | **4** failures |

The last pair is the one that matters: the gate-7 suite pins both edges, so neither a mute nor a revert can pass it.

## The runner corrupted its own measurements

61 gates wrote to hardcoded `/tmp/hydra-gate-<name>.log` and derived their verdicts by `wc -l` on them. Exactly one used `mktemp`, and its comment said why.

Two concurrent runs on **different repos**, v1.4.0:

```
petstore:     [gate-46] FAIL — 26 unresolved @spec target(s)
              — see /tmp/hydra-gate-spec-anchor-existence.log   ← 0 lines
app-versions: [gate-46] PASS
petstore:     [gate-40] FAIL — 1     ┐ one file,
app-versions: [gate-40] FAIL — 7     ┘ 7 lines
```

petstore's 26 findings were truncated away by app-versions. **Had the truncation landed before the `wc -l` instead of after, petstore would have reported PASS over 26 real findings.** Same test on this branch: two private directories, each containing exactly what its own run reported (26/0 and 0/1).

The package's own test suite had it too — `test_gate_route_auth.sh` reported **7 failures under the harness and 0 standalone**. Fixed there as well.

## A resolving diff base is not a usable one

shillinq's `development` run finished in **22 seconds, all green**. Not a shallow checkout: on a push to a mainline branch `origin/development` **IS** `HEAD`, so the diff is empty by construction and every gate passes over nothing. Verified at `c64e9fe` — **52 gates PASS scoped, 18 FAIL unscoped**.

Now refused (exit 99), with a merge-base check for genuine shallow checkouts, and the diff's exit code read **directly** rather than through a `||` chain that cannot tell "no changes" from "could not run".

⚠️ **This will turn currently-green fleet runs red.** They are falsely green. Worth knowing before the roll.

## gate-28: the NUL byte is worse than #171 filed

Depending on the grep implementation a raw `0x00` produces **either** outcome:

- GNU grep ≤3.4 prints `Binary file X matches` on **stdout** → `awk '{print $3}'` reads the **file path** as the licence → false RED
- ugrep / GNU grep ≥3.5 print nothing to stdout → the gate `continue`s and **never checks the file** → false GREEN

Verified: an `@license AGPL-3.0-or-later` hidden behind a NUL **passed silently**. Live NULs remain in `doriath/src/import/model.js` and `openbuild/src/services/manifestValidation/documentAttachments.js` (`file` reports both as `data`).

The gate also read only the first `@license` and ignored `SPDX-License-Identifier:` entirely — how 174 files carried an AGPL claim behind a green gate. Now every declaration is collected; identifiers inside string literals stay test data (nldesign's `MarianneFontTest.php`).

## gate-9: the advice would have introduced the vulnerability

> *"remove `#[PublicPage]` or remove body auth check"*

The first breaks the endpoint — NC middleware rejects the remote caller before the controller runs. The second deletes its only authentication. **There was no correct action a developer could take on 34 of 45 findings.** Returning 401/403 is not, on its own, evidence of a session dependency; it is what a self-authenticating public endpoint does.

A `#[PublicPage]` method that tests the **session** still fires, under a rule name that says so, with advice that never tells anyone to open an endpoint. Asserted on the string itself, in `RemediationTextIsSafe`.

## gate-7: anti-correlated with the property it checks

On a multi-tenant codebase the tenancy guard refuses with **404 on purpose**, because a 403 leaks another tenant's object ids. gate-7 excluded bare throws, so it flagged exactly the code that got tenancy right — and `FlowController::state()` reported **identically before and after its real IDOR was fixed**.

Adds a signal requiring **both** a comparison against a session-derived scope **and** a refusal. **Partial**: 6 of ~17 openregister false positives clear. The rest need collaborator-hop work — left for a follow-up rather than guessed at, because this is a security gate.

## Also

- **19 helper lookups** re-resolved `dirname "${BASH_SOURCE[0]}"` *after* the `cd "${APP_DIR}"`, against the warning at the top of the file. One (gate-17) **aborted the whole suite** when the runner was invoked by a relative path — 46 gates never ran.
- Gates 46/40/28 now `_skip(wiring)` when their helper is missing rather than passing over an unread file set (#147).
- gate-40 is ~40× faster: one python process, not one per `.vue`.
- `quality.yml` **floats on `@main` while the package it drives is PINNED**, and executes paths inside it *by name*. That interface broke three times in one day (#168). Pinning both halves from one tag is a **human call** — it changes how 22 repos receive gate fixes. Until then a preflight names the desync instead of letting it surface as an unexplained gate failure. Largest single `run:` step in the file is 8,533 B, well under the ~19 KB limit.

## Verification

```
$ bash hydra-gates/tests/run-helper-suites.sh
   passed: 22   quarantined: 2   failed: 0
```
22 suites green (v1.4.0: 17 passing, 1 failing). 3 concurrent harness runs: 0 failures each — on v1.4.0 one of two overlapping runs failed.

## Release

Ready to tag **v1.5.0**, not v1.4.x. New refusal conditions (exit 97, exit 99 on base==HEAD) turn previously-passing runs into failures, and gates emit new rule names. That must not arrive as a patch. The fleet pins **v1.3.0**, so anyone bumping to a v1.4.x would silently also receive v1.4.0's unrolled `gh_slugify` verdict change — one minor makes the whole delta legible in one release note.

**Not rolled to the fleet.** Reported ready.

Closes #158 (items 2 and 6), closes #171. Partially addresses #160 (items 2 and 3; item 1 partial). Addresses #159's `@main`-vs-pinned split with a named preflight, not a resolution.

### DIFF
--- a/.github/workflows/hydra-gates-package.yml
+++ b/.github/workflows/hydra-gates-package.yml
@@ -252,16 +252,31 @@ jobs:
           # The offending FILES must be named, so we know the gates fired on
           # what we injected and not on something else that happened to be
           # there. stdout carries only the COUNT per gate by design; the file
-          # list goes to the gate's own /tmp/hydra-gate-<name>.log, so that is
-          # where the attribution has to be checked. A gate that reported FAIL
-          # while writing no evidence at all would itself be a defect, so a
-          # missing log is a failure here too.
+          # list goes to the gate's own log, so that is where the attribution
+          # has to be checked. A gate that reported FAIL while writing no
+          # evidence at all would itself be a defect, so a missing log is a
+          # failure here too.
+          #
+          # The log DIRECTORY is read from the run's own first line rather than
+          # hardcoded. It used to be the shared /tmp/hydra-gate-<name>.log, and
+          # that was the same defect this assertion exists to catch, one level
+          # up: two runs on one host overwrote each other's evidence, so this
+          # check could read a DIFFERENT run's log and pass, or read a truncated
+          # one and fail, without either outcome saying anything about the
+          # fixture. Each invocation now mints its own directory and announces
+          # it; asking the run where it wrote is the only way to be sure the
+          # evidence belongs to it.
+          LOGDIR="$(printf '%s\n' "${OUT}" | sed -n 's/^\[hydra-gates\] findings logs: //p' | head -1)"
+          if [ -z "${LOGDIR}" ] || [ ! -d "${LOGDIR}" ]; then
+            echo "::error::the run did not announce a findings-log directory (expected a '[hydra-gates] findings logs: <dir>' line). Without it there is no way to attribute a verdict to its evidence."
+            exit 1
+          fi
           for probe in "forbidden-patterns:lib/GateControl.php" "window-confirm:src/gateControl.js"; do
             gate="${probe%%:*}"; want="${probe##*:}"
             # shellcheck disable=SC2012
-            log="$(ls -t "/tmp/hydra-gate-${gate}.log" /tmp/hydra-gate-"${gate}".*.log 2>/dev/null | head -1)"
+            log="$(ls -t "${LOGDIR}/hydra-gate-${gate}.log" "${LOGDIR}"/hydra-gate-"${gate}".*.log 2>/dev/null | head -1)"
             if [ -z "${log}" ]; then
-              echo "::error::gate ${gate} reported a result but wrote no /tmp/hydra-gate-${gate}*.log — no evidence for its verdict"
+              echo "::error::gate ${gate} reported a result but wrote no ${LOGDIR}/hydra-gate-${gate}*.log — no evidence for its verdict"
               exit 1
             fi
             echo "--- ${log}"; cat "${log}"
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -3178,6 +3178,52 @@ jobs:
           ref: ${{ inputs.hydra-gates-ref }}
           path: gates
 
+      # ── the floating-caller / pinned-callee contract ──────────────────────
+      #
+      # THIS WORKFLOW FLOATS ON @main; THE PACKAGE IT DRIVES IS PINNED.
+      #
+      # A caller references this file as
+      # `ConductionNL/.github/.github/workflows/quality.yml@main` while
+      # pinning `hydra-gates-ref` to a tag. So the two halves of one system
+      # move independently, and every path this workflow executes inside
+      # `gates/hydra-gates/` is an UNVERSIONED INTERFACE between them.
+      #
+      # That interface has broken three times in one day. The decisive one
+      # was #168: a change on @main made this workflow execute
+      # `gates/hydra-gates/scripts/axe-run.cjs` BY PATH, and every repo whose
+      # pin predated that file got a failure whose message named neither the
+      # pin nor the file.
+      #
+      # Pinning both halves from one tag, or floating both, is the real fix
+      # and is a human decision — it changes how 22 repos receive gate fixes.
+      # Until then this step makes the desync SAY SO. A named refusal is not
+      # a fix, but it is the difference between "gate-33 is broken" and
+      # "your pin is four fixes behind and here is the file it is missing".
+      - name: Verify the pinned gates package satisfies this workflow
+        run: |
+          set -uo pipefail
+          REF='${{ inputs.hydra-gates-ref }}'
+          MISSING=""
+          for p in \
+            bin/hydra-gates \
+            scripts/run-hydra-gates.sh \
+            scripts/axe-run.cjs \
+            scripts/lib/check_spec_anchors.py \
+            scripts/lib/check_form_labels.py \
+            scripts/lib/check_license_triangle.py \
+            scripts/lib/check_no_admin_idor.py \
+            scripts/lib/check_semantic_auth.py \
+            scripts/lib/check_orphan_auth.py
+          do
+            [ -e "${GITHUB_WORKSPACE}/gates/hydra-gates/${p}" ] || MISSING="${MISSING} ${p}"
+          done
+          if [ -n "${MISSING}" ]; then
+            echo "::error::hydra-gates-ref '${REF}' does not contain:${MISSING}"
+            echo "::error::This workflow floats on @main and executes those paths by name inside the PINNED package, so a pin older than them cannot run the gates they implement. Bump hydra-gates-ref to a tag that includes them. This is NOT a code-quality finding about your repository."
+            exit 1
+          fi
+          echo "Pinned gates package at '${REF}' provides every path this workflow executes."
+
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
--- a/hydra-gates/scripts/lib/check_form_labels.py
+++ b/hydra-gates/scripts/lib/check_form_labels.py
@@ -1,0 +1,249 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: EUPL-1.2
+"""Gate-40 form-label-association — every form control must have an
+accessible name (WCAG 2.2 AA, SC 1.3.1 Info and Relationships and SC 3.3.2
+Labels or Instructions).
+
+WHY THIS WAS REWRITTEN
+----------------------
+The previous implementation flattened the file's newlines into spaces and
+ran four independent regexes over the result. It had no notion of nesting,
+of a component's slot, or of where the template ends — so it could only see
+attributes ON the element, and an accessible name that comes from anywhere
+else read as an absent one. Measured across 21 fleet repos: **1,211
+findings, 58% of them false**, in four shapes:
+
+  implicit label wrapping   `<label><input …><span>Safe mode</span></label>`
+                            is the canonical HTML association and needs no
+                            `for`/`id` pair at all. 268 findings.
+
+  NcCheckboxRadioSwitch     `<NcCheckboxRadioSwitch v-model="x">Installed
+  default slot              apps only</NcCheckboxRadioSwitch>` — nc-vue
+                            renders the default slot INTO the `<label>`.
+                            463 findings, and this is the dangerous one:
+                            the only way to satisfy the old gate was to add
+                            `aria-label`, which OVERRIDES the visible label
+                            and breaks speech-input users who say what they
+                            see. A gate whose remediation is an
+                            accessibility REGRESSION cannot be closed
+                            honestly.
+
+  dynamic :id / :for        `<label :for="`f-${id}`">` + `<input
+                            :id="`f-${id}`">` is a correct association the
+                            literal-only regex could not see. 56 findings.
+
+  commented-out markup      `<!-- <input type="text"> -->` ships nothing.
+                            5 findings.
+
+WHAT STILL FAILS, AND MUST
+--------------------------
+A control with no accessible name from ANY of the recognised sources is
+still reported. Every relaxation here is anchored to a real naming
+mechanism the browser implements — a wrapping `<label>`, a matching
+`for`/`id` pair (literal or expression), a component prop the library
+documents, or slot content the component renders into its own `<label>`.
+An `<input type="text">` standing on its own, a self-closed
+`<NcCheckboxRadioSwitch />` with no prop and no slot, and a `<label>` whose
+`for` matches no control are all still findings. See
+``test_check_form_labels.py``, where each relaxation ships with the
+true-positive case it must not swallow.
+
+Usage:
+    check_form_labels.py <file.vue>...        # findings on stdout
+"""
+from __future__ import annotations
+
+import re
+import sys
+
+# Native controls that need a name. `select` is deliberately absent: it is
+# gate-12 (nc-input-labels)' subject, and widening this gate's remit while
+# fixing its false-positive rate would make the two numbers incomparable.
+NATIVE = {'input', 'textarea'}
+# Nextcloud components whose published API takes the name as a prop.
+# `NcSelect` belongs to gate-12, same reason.
+NC_PROP_COMPONENTS = {'nctextfield', 'ncinputfield', 'ncrichcontenteditable',
+                      'nccheckboxradioswitch'}
+# ...of which these ALSO accept the name as default-slot content, because
+# they render that slot inside their own <label> element.
+NC_SLOT_COMPONENTS = {'nccheckboxradioswitch'}
+
+# `type=` values that carry their own name or take none.
+EXEMPT_INPUT_TYPES = {'hidden', 'submit', 'button', 'reset', 'image'}
+
+TAG = re.compile(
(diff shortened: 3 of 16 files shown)


rubenvdlinde:
### Update — rebased onto `main`, reconciled with #172/#173/#174

`main` moved three commits while this was in flight, two of them in the same
gate. Reconciled rather than overwritten:

- **#172** gave gate-28 a four-way skip taxonomy and made PASS conditional on
  having compared at least one file. Moving the read into a helper had
  collapsed all of that into one `na` — **undoing #172**. Taxonomy restored
  verbatim; the helper now returns the compared-file count on stderr so
  `_lt_checked` still gates the PASS. The one state that short-circuits it is
  a **missing helper**, which #172's chain would have called `structural` — a
  claim about the *repository*, when the repository is fine and the gate is
  broken.
- **#173** added a preflight for the **flag** direction of the
  floating-caller/pinned-callee desync. This PR's preflight covers the **path**
  direction (#168's shape). Both kept — different halves of one interface.
- Gates 40 and 46 were declaring `na` on an empty in-scope set. That is
  ordinary ADR-020 diff scoping, and `test-hydra-gates-bin.sh` asserts no
  src-guarded gate goes NOT APPLICABLE while `src/` exists. Reverted to the
  sibling behaviour.

**One contract change, now explicit.** `test-hydra-gates-bin.sh`'s *"empty diff
exits 0"* fixture set `BASE_SHA` to `HEAD` — so it was really asserting
*"scoping a commit against itself exits 0"*, which is the shillinq shape. Those
are different facts and only one is legitimate. The original assertion now runs
against a real base-behind-HEAD empty diff and still expects 0; a **new**
assertion covers `base == HEAD`, expects 99, and checks that **no** gate printed
PASS.

```
hydra-gates entry-point tests: 31 passed 3 failed  ->  36 passed 0 failed
helper suites:                 23 passed 0 failed
gate-46 fleet 1,995 -> 918     gate-40 fleet 1,211 -> 517   (unchanged by the merge)
```

Also fixed a CodeQL `py/bad-tag-filter` (high) that this PR introduced in its
own terminator-trimming regex: it knew `-->` but not `--!>`, so
`<!--SPDX-License-Identifier: EUPL-1.2--!>` would have yielded `EUPL-1.2--!` as
the licence and reported drift on a correct header. Tested both ways.

PR #185, opened 2026-08-06
fix(gate-64): a named constant is the prelude, spelled better — not a missing one

rubenvdlinde:
## The finding that had nothing behind it

gate-64 recognised the ADR-040 autoload prelude only as a **quoted literal inside the call parentheses**:

```php
\OC_App::registerAutoloading('openregister', $path);              // accepted
\OC_App::registerAutoloading(self::OPENREGISTER_APP_ID, $path);   // REJECTED
```

The second is the same call with the app id given a name, four lines below `private const OPENREGISTER_APP_ID = 'openregister';`. doriath writes it that way — and gate-64 reported it as an AppHost adoption with **no** prelude, with the prelude present, correct, and unit-tested.

Measured on doriath#163: `FAIL — 2 AppHost adoption(s) with no OpenRegister autoload prelude`.

A gate that fails the tidier spelling of the thing it is asking for is not reporting a defect. It teaches people to write the untidy spelling, and it spends the reader's attention on nothing.

## The fix, and why it does not go blind

Resolution is deliberately **shallow and literal**: a name counts only when the same source blob binds it to the exact string `'openregister'` — via `const`, `define()`, or a plain variable. Imports are not followed, and **a name that cannot be seen bound is not accepted**.

That last property is the whole design. "Accept any argument" would clear doriath and blind the gate in the same stroke, so two of the four new tests exist to fail in exactly that direction.

| test | direction |
|---|---|
| `test_named_constant_bound_to_openregister_is_GREEN` | doriath's shape passes |
| `test_variable_and_define_spellings_are_accepted` | the other two bindings |
| `test_constant_bound_to_another_app_still_FAILS` | `registerAutoloading()` for a **different** app is not a prelude |
| `test_unknown_constant_still_FAILS` | an unresolvable name is not taken on trust |

### Mutants — disjoint sets, both directions

| mutant | tests turned red |
|---|---|
| accept any argument name | the 2 discriminators |
| resolve nothing (regression to the bug) | the 2 acceptance tests |

## Measured against the real trees

This is what settles it — the gate must lose the false positive **and keep both true positives**.

| repo | gate-64 @ `main` | with this fix | |
|---|---|---|---|
| doriath | `rc=1` FAIL | `rc=0` PASS | false positive cleared |
| scholiq | `rc=1` FAIL | `rc=1` FAIL | real defect, still caught |
| openbuild | `rc=1` FAIL | `rc=1` FAIL | real defect, still caught |

scholiq and openbuild genuinely have no prelude, and openbuild is the acute one: its `class_exists(Bootstrap::class)` guard answers **false** on a perfectly healthy instance, because `openbuild` sorts before `openregister` and `Coordinator::registerApps()` registers autoloading one app at a time. So `Bootstrap::register()` has apparently never run there, and its generic controllers, install repair steps and deep-link listener are silently absent. Those are fixed in their own repos, not here.

⚠️ Note on my own method: my first before/after table was **wrong in the safe-looking direction** — I copied the "baseline" helper *after* patching it, so both columns ran the fixed matcher and doriath appeared to pass on `main`. The table above is re-measured against `git show origin/main:` and the md5sums were checked. A baseline captured after the edit is not a baseline.

18 gate-64 tests pass.

### DIFF
--- a/hydra-gates/scripts/lib/check_apphost_autoload_prelude.py
+++ b/hydra-gates/scripts/lib/check_apphost_autoload_prelude.py
@@ -93,6 +93,34 @@ import sys
 PRELUDE = re.compile(r"registerAutoloading\s*\(([^)]*)\)", re.S)
 OPENREGISTER_LITERAL = re.compile(r"""['"]openregister['"]""")
 
+# A name BOUND to the literal 'openregister' in the same blob. Three spellings:
+#   const OPENREGISTER_APP_ID = 'openregister';
+#   define('OPENREGISTER_APP_ID', 'openregister');
+#   $openRegisterAppId = 'openregister';
+# The binding must be to the EXACT literal — this resolves names, it does not
+# guess at them, and a name bound to anything else is not a match.
+CONST_BINDING = re.compile(
+    r"""(?:
+          const\s+(?P<name1>[A-Za-z_][A-Za-z0-9_]*)\s*=\s*['"]openregister['"]
+        | define\s*\(\s*['"](?P<name2>[A-Za-z_][A-Za-z0-9_]*)['"]\s*,\s*['"]openregister['"]
+        | \$(?P<name3>[A-Za-z_][A-Za-z0-9_]*)\s*=\s*['"]openregister['"]
+    )""",
+    re.X,
+)
+
+
+def _binding_name(match: "re.Match[str]") -> str:
+    return match.group("name1") or match.group("name2") or match.group("name3")
+
+
+# A name USED as a call argument: `self::NAME`, `static::NAME`, `Cls::NAME`,
+# a bare `NAME`, or `$var`. Matched against the bound names above, never
+# accepted on its own.
+ARG_NAME = re.compile(
+    r"""(?:(?:self|static|parent|[A-Za-z_][A-Za-z0-9_\\]*)\s*::\s*)?"""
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
Superseded by #184, which landed while this was in review and is a **superset** of it.

Both fixes resolve the same false positive — gate-64 matching the app id only as a quoted literal, so doriath's `registerAutoloading(self::OPENREGISTER_APP_ID, $path)` read as a missing prelude. #184 additionally strips comments before matching, which closes a **false GREEN** this PR did not address: a commented-out `registerAutoloading('openregister', …)` would have counted as a prelude that no longer runs. That is the more dangerous direction and it belongs in the gate.

Re-measured against the real trees with #184 as it now stands on `main`:

| repo | gate-64 |
|---|---|
| doriath | `rc=0` PASS — false positive cleared |
| scholiq | `rc=1` FAIL — real defect, still caught |
| openbuild | `rc=1` FAIL — real defect, still caught |

Identical to this PR's outcome, so nothing is lost by closing it.

The only delta here was accepting two further bindings — `define('NAME', 'openregister')` and `$var = 'openregister'` — where #184 accepts `const` only. No repo in the fleet uses either spelling today, so I am not re-opening it as a follow-up; if one ever does, gate-64 will say so by name and the binding can be added then with its own paired test.

The two real defects this exposed are being fixed in their own repos: ConductionNL/scholiq#289 and ConductionNL/openbuild#139.

Issue #205, opened 2026-08-07
Every reason-bearing gate opt-out is unreachable on a PR: HYDRA_GATE_PR_BODY is never set, and git log -1 reads the merge commit


rubenvdlinde:
Several gates document a reason-bearing escape hatch. None of them can fire on a `pull_request` run.

## The two paths, and why both are dead

`run-hydra-gates.sh` (v1.5.0 and main, identical):

```sh
[ -n "${HYDRA_GATE_PR_BODY:-}" ] && echo "${HYDRA_GATE_PR_BODY:-}" | grep -qE "${_cxt_optout_re}" && _cxt_optout="1"
[ -z "${_cxt_optout}" ] && git log -1 --pretty=%B 2>/dev/null | grep -qE "${_cxt_optout_re}" && _cxt_optout="1"
```

1. **`HYDRA_GATE_PR_BODY` is never exported.** `grep -n HYDRA_GATE_PR_BODY .github/workflows/quality.yml` returns **nothing** on `main`. The variable the first path depends on is always empty.
2. **`git log -1` reads the wrong commit.** On a `pull_request` event `actions/checkout` checks out the synthetic **merge** commit, whose message is `Merge <sha> into <sha>` — never the author's.

So the documented opt-out works only on a direct push to a mainline branch, which is the case where it is least needed.

## Measured

On ConductionNL/hermiq#162 I put the exact tag in **both** places:

- in the PR body (`gh pr edit --body-file`), then re-ran the failed jobs — gate-49 still FAIL
- in the head commit message, as an explicit empty commit (`0d095f24`), pushed — gate-49 still FAIL

Same finding count each time.

## Affected gates (at least)

- `gate-49` controller-exception-translation — `run-hydra-gates.sh:3562`
- `schema-change-has-tests` — `:3398`
- `csrf-cochange` — `:3448`

Any gate whose only escape hatch is this mechanism is effectively **non-overridable on a PR**, however well-justified the reason.

## Suggested

Export it in `quality.yml`, which already has the PR context:

```yaml
env:
  HYDRA_GATE_PR_BODY: ${{ github.event.pull_request.body }}
```

and, for the fallback, read `${{ github.event.pull_request.head.sha }}` rather than `git log -1`, so the author's own commit message is what is inspected.

## Why it matters now

[...]

Issue #204, opened 2026-08-07
gate-49 does not recognise catch (\\Throwable), which is strictly broader than the nine it tracks


rubenvdlinde:
gate-49 accepts a `catch` only for one of nine named domain exceptions:

```
TRACKED = [DoesNotExistException, MultipleObjectsReturnedException, NotFoundException,
           PermissionException, ValidationException, ForbiddenException,
           CustomValidationException, AppendOnlyException, ArchivalImmutableException]
```

`catch (\Throwable $e)` matches none of them — yet it catches **all nine, plus everything else**. So the broadest possible translation is reported as no translation at all.

Hit on ConductionNL/hermiq#162: `SettingsController::index/create/load` were flagged with no handling, I added `catch (\Throwable)` to each with a translated JSON response and a log, and the gate still reports "3 controller method(s) missing try/catch or @throws".

## Why it matters beyond a false positive

The finding pushes authors toward the narrower fix. Catching only `DoesNotExistException` where the real risk is "the service blew up for a reason I have not enumerated" leaves the uncaught path — a framework 500 with a stack trace — exactly where it was. For `index()` that trace reaches a **non-admin**, since the method is `#[NoAdminRequired]`.

## Suggested

Treat `catch (\Throwable)` and `catch (\Exception)` as satisfying the gate. They are supersets of the tracked list; a gate that rejects the stronger guarantee teaches people to write the weaker one.

Worked around for now with the documented opt-out (`[hydra-gate-controller-exception-translation exclude] …`), which is the right escape hatch but should not be needed for the broadest correct handler.

Related: #203 — gate-55 and gate-60 encode two different icon registries for the same widget field, with no opt-out at all.

rubenvdlinde:
Fixed in ConductionNL/.github#211 (merged 2026-08-07).

Each fix ships with a control proving the gate can still fail:
- gate-49 now matches `\Throwable`/`\Exception`; no-catch and an unrelated class still do not
- `_optout_text` reads the PR body and the author's head SHA; verified that plain `git log -1` does NOT see a tag written behind a merge commit while the helper does

Note for #204 specifically: the fix is inside the PINNED hydra-gates package, so it reaches a repo only when its `hydra-gates-ref` moves to a tag containing it. Tag + consumer sweep to follow.

PR #211, opened 2026-08-07
fix(gates): make the opt-outs reachable, stop rejecting \Throwable, end the icon deadlock

### LINKED ISSUE #203: gate-55 and gate-60 encode two different icon registries for the same widget field — 'settings' is unsatisfiable
A widget icon for the "settings" concept cannot satisfy both gates. Verified against the installed library, not inferred.

## The deadlock

| gate | rule | verdict on `Cog` | verdict on `CogOutline` |
|---|---|---|---|
| **55** detail-page-discipline | widget `icon` must be in the shared widget registry | PASS | FAIL — "not in the shared icon registry (renders the ? fallback, ADR-062 rule 8)" |
| **60** icon-vocabulary | concept `settings` MUST use `CogOutline` (ADR-077 Tier A) | FAIL | PASS |

Hit on ConductionNL/hermiq#162, widget `agent-core` titled "Configuration" on `AgentDetail`.

## Which registry is right

Checked the source of truth `check_detail_page_discipline.py` names — `nextcloud-vue/src/components/CnWidgetGrid/widgetIcons.js`, as installed in `node_modules`:

```
grep -c CogOutline widgetIcons.js   ->  0
grep -c bCogb   widgetIcons.js   ->  2
```

So gate-55's hardcoded mirror is **accurate**: `CogOutline` genuinely is not there, and using it renders the `?` fallback. ADR-077 Tier A therefore mandates, for widget icons, a glyph the widget renderer cannot draw.

## Why this is gate-60's overreach, not a repo defect

Gate 60 is introduced in `run-hydra-gates.sh:3908` as *"every manifest **menu** `icon`"*, but `_menu_entries()` deliberately walks the whole tree:

> Originally menu-only. … page/tab/widget icons and `actions[]` / `headerActions[]` entries render through the same CnIcon registry and hit exactly the same failure modes … So the gate now walks the whole tree.

That premise does not hold for **widget** icons: they render through `CnWidgetGrid`'s own `widgetIcons.js` registry, which is a strict subset of the CnIcon vocabulary. Two registries, one field.

## Suggested

Either
1. exempt widget icons from gate-60's Tier A concept mapping (menus keep it), or
2. reconcile the registries — add `CogOutline` and the other Tier A glyphs to `widgetIcons.js`, so the vocabulary ADR-077 mandates is one the widget renderer can actually draw.

(2) is the better end state; (1) unblocks immediately.

Until then, any PR touching a widget labelled Configuration / Settings / Instellingen / Configuratie is blocked, and gate-60 has **no documented opt-out** — unlike gate-49, which takes `[hydra-gate-controller-exception-translation exclude] <reason>`.

## Related

gate-49 has a narrower version of the same shape: it tracks nine named domain exceptions and does not recognise `catch (\Throwable)`, which is strictly BROADER and covers all nine. Filing separately.

rubenvdlinde:
Closes #203, #204, #205 — three defects found while landing ConductionNL/hermiq#162, where the gates blocked a PR on findings its author could neither fix nor waive.

Each fix ships with a **control proving the gate can still fail**. A gate that cannot fail is worth as little as one that always does.

## 1. Every reason-bearing opt-out was unreachable on a PR (#205)

The opt-outs read `${HYDRA_GATE_PR_BODY}`, falling back to `git log -1`. On a `pull_request`, **both read nothing**:

- the variable was never exported by `quality.yml`, so the first path was always empty
- `actions/checkout` checks out the synthetic **merge** commit (`Merge <sha> into <sha>`) — never a message an author wrote

Measured on hermiq#162: the tag was placed in the PR body **and** in an explicit head commit; the finding count did not move either time.

`quality.yml` now exports `HYDRA_GATE_PR_BODY` and `HYDRA_GATE_HEAD_SHA`, and the three call sites go through one `_optout_text` helper — so they cannot drift apart, and a fourth gate adding an opt-out inherits the fixed behaviour instead of copying the broken pair of lines.

**Control** — a real author commit behind a merge commit:

```
git log -1        -> tag NOT found   (the bug)
_optout_text      -> tag found       (fixed)
PR body only      -> tag found
```

## 2. gate-49 rejected `catch (\Throwable)`, which is strictly broader (#204)

It matched nine named domain exceptions. `\Throwable` and `\Exception` are **supersets of all nine**, so the broadest possible translation was reported as no translation at all.

The cost is not the red check — it is that the finding pushes the author toward the **narrower** handler: catch the one named exception, leave everything else to become a framework 500 with a stack trace. For a `#[NoAdminRequired]` method that trace reaches a non-admin. A gate that rejects the stronger guarantee teaches people to write the weaker one.

**Control** — `\Throwable`, `\Exception`, `DoesNotExistException` now match; no-catch and an unrelated class still do not.

## 3. gate-55 and gate-60 demanded different icons for the same field (#203)

ADR-077 Tier A requires `CogOutline` for the `settings` concept. The widget renderer's own registry — `CnWidgetGrid/widgetIcons.js`, which gate-55 enforces — ships `Cog` and **not** `CogOutline`. Verified against the installed library rather than inferred:

```
grep -c CogOutline widgetIcons.js  -> 0
grep -c bCogb   widgetIcons.js  -> 2
```

So obeying gate-60 on a widget renders the `?` fallback, and obeying gate-55 fails gate-60. Unsatisfiable — and gate-60 has **no opt-out at all**.

Widget icons are now exempt from the concept MUST **and from nothing else**: a nonexistent icon or an unbridged `icon-*` still fails wherever it appears, and menus keep the full Tier A vocabulary.

**Control** — a MENU labelled "Settings" carrying `Cog` still **FAILS**; the same value on a widget passes. The previously-deadlocked hermiq manifest goes 1 failure → 0.

The better end state is reconciling the two registries (add the Tier A glyphs to `widgetIcons.js`), after which this exemption can go. Until then it is the difference between a gate that is strict and one that is impossible.

## Rollout note

Fix 1 touches `quality.yml`, which every caller consumes `@main` — it takes effect fleet-wide on merge. It is purely additive (two new env vars), so an older pinned gates package simply ignores them.

Fixes 2 and 3 are inside the pinned `hydra-gates` package, so they reach a repo only when its `hydra-gates-ref` moves to a tag containing them. Worth a tag and a consumer sweep, as with v1.5.0.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -3546,6 +3546,26 @@ jobs:
         id: gates
         env:
           HYDRA_GATE_BASE_REF: ${{ steps.base.outputs.base }}
+          # Several gates document a reason-bearing opt-out —
+          # `[hydra-gate-<name> exclude] <reason>` — and read it from THIS
+          # variable, falling back to `git log -1`. Neither reached anything:
+          #
+          #   - this variable was never set, so the first path was always empty
+          #   - on a pull_request, actions/checkout checks out the synthetic
+          #     MERGE commit, whose message is `Merge <sha> into <sha>`, so the
+          #     fallback inspected a message no author wrote
+          #
+          # So every reason-bearing opt-out in the suite was unreachable on a
+          # PR — the one place they are needed. Measured on hermiq#162: the tag
+          # was placed in the PR body AND in an explicit head commit, and the
+          # finding count did not move either time.
+          #
+          # `github.event.pull_request.body` is null on a push, which is
+          # correct: the fallback then reads the real pushed commit.
+          HYDRA_GATE_PR_BODY: ${{ github.event.pull_request.body }}
+          # The author's own head commit, so the fallback stops reading the
+          # merge commit. Empty on a push, where `git log -1` is already right.
+          HYDRA_GATE_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
           set -u
           ARGS="--app-dir ${GITHUB_WORKSPACE}/app"
--- a/hydra-gates/scripts/lib/check_icon_vocabulary.py
+++ b/hydra-gates/scripts/lib/check_icon_vocabulary.py
@@ -178,7 +178,7 @@ def _menu_entries(path: str):
     except (OSError, ValueError):
         return
 
-    def walk(node, label_ctx):
+    def walk(node, label_ctx, in_widget=False):
         if isinstance(node, dict):
             if node.get('type') == 'caption':
                 return
@@ -186,13 +186,16 @@ def _menu_entries(path: str):
                      or node.get('id') or label_ctx)
             icon = node.get('icon')
             if isinstance(icon, str) and icon:
-                yield (str(label or ''), icon, str(node.get('id') or ''))
+                yield (str(label or ''), icon, str(node.get('id') or ''), in_widget)
             for key, value in node.items():
                 if key != 'icon':
-                    yield from walk(value, label)
+                    # Once inside a `widgets` array, everything below it is a
+                    # widget icon and stays flagged as one. See the Tier A
+                    # concept check for why that distinction matters.
+                    yield from walk(value, label, in_widget or key == 'widgets')
         elif isinstance(node, list):
             for item in node:
-                yield from walk(item, label_ctx)
+                yield from walk(item, label_ctx, in_widget)
 
     yield from walk(data, '')
 
@@ -255,7 +258,7 @@ def main() -> int:
 
     for path in paths:
         rel = os.path.relpath(path, repo)
-        for label, icon, entry_id in _menu_entries(path):
+        for label, icon, entry_id, in_widget in _menu_entries(path):
             where = f'{rel}: {label or entry_id or "?"}'
             if not icon:
                 continue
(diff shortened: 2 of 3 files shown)


rubenvdlinde:
Correction to the rollout note in the description, checked after merging.

I wrote that fixes 2 and 3 sit in the pinned package and would need "a tag and a consumer sweep". **That is no longer true.** The fleet has since dropped `hydra-gates-ref` entirely — every repo I checked (openregister, opencatalogi, openconnector, docudesk, procest, pipelinq, openbuild, nldesign) now has no pin and tracks `@main`, with a comment in each `code-quality.yml` explaining why:

> A pin is a silent expiry date: 22 repos sat on v1.0.1 and 16 gates were dead fleet-wide while every one reported PASS (#159), and a default flipped at @main later reached those old runners and made them red on gates they had no subject matter for (#173).

So **all three fixes reach every repo immediately** — no sweep required, and the earlier v1.5.0 sweep is superseded by the better arrangement.

I did cut `v1.6.0` at the merge commit before establishing this. It is harmless and verified to carry both package-side fixes (`_optout_text` ×5, `in_widget` ×6), so it remains available for anyone who deliberately pins with the documented `hydra-gates-ref` input. Nothing depends on it.

Worth recording: my first pin survey returned "none" for all 15 repos and I assumed my decode had broken. It had not — the data had changed. Checking before acting on that assumption is what surfaced the arrangement.

PR #219, opened 2026-08-08
fix(gate-12): the element ended at the arrow of `:reduce="(o) => o.id"` — 18 of 18 findings false on scholiq

rubenvdlinde:
fix(gate-12): the element ended at the arrow of `:reduce="(o) => o.id"`

Gate-12 extracted `<NcSelect>` elements with

    grep -oE '<NcSelect[^>]*>'

`[^>]*` stops at the FIRST `>` character in the flattened source. In an
NcSelect that is almost never the end of the tag, because the idiomatic
vue-select usage puts an arrow function in an attribute value:

    <NcSelect
        v-model="selectedRoundId"
        :options="roundOptions"
        :reduce="(o) => o.id"
        :input-label="t('scholiq', 'Round')"
        :aria-label-combobox="t('scholiq', 'Round')" />
                  ^ extraction ends here

Everything after `:reduce` is invisible, so the two props the gate looks
for are cut off and the element is reported as unnamed.

MEASURED, scholiq 2026-08-08 full-tree: **18 findings, 18 of them false.**
Every flagged element already carried `:input-label` AND
`:aria-label-combobox`, in every case written after `:reduce`. A
quote-aware re-scan of the same nine files found 18 NcSelect elements and
18 with a label prop.

The gate was anti-correlated with its own subject in those files: adding
the label prop could not clear it, and deleting `:reduce` could. Same
shape as gate-9's `[^)]*` bug (#198) — a character class used as a
delimiter over content that legitimately contains the delimiter.

FIX
---
Extraction moves to `scripts/lib/check_nc_select_labels.py`, using the same
quote-aware tag pattern `check_form_labels.py` (gate-40) already uses, so
the two a11y gates now agree on where an element stops. Wired with the
missing-helper `_skip` pattern from #147: an absent helper reports
SKIPPED (wiring) naming how many components went uninspected, never PASS.

The accepted prop set is UNCHANGED — `input-label` / `inputLabel` /
`aria-label-combobox` / `ariaLabelCombobox`, bare or bound. This changes
where an element ENDS, not what counts as a name, so before/after numbers
stay comparable. `aria-label` is still not accepted; a manual `<label
for=…>` next to an NcSelect is still a finding.

Markup inside an HTML comment is no longer scanned — a commented-out
NcSelect renders nothing and can carry no accessible name.

EVIDENCE
--------
- 17 new unit tests in `test_check_nc_select_labels.py`, discovered
  automatically by `tests/run-helper-suites.sh`. Every relaxation is
  paired with the true positive it must not swallow:
  `:reduce` + label -> clean, the same element with the label deleted ->
  1 finding; two selects where only one is named -> exactly 1 finding
  (the greedy-match trap a naive fix falls into); an unclosed `<NcSelect`
  followed by a `<div>` -> 1 finding that does not name the div.
- Whole package suite: 28 passed, 0 failed, 2 pre-existing quarantines.
- Against scholiq's tree, unchanged app code: gate-12 goes 18 -> PASS.
- Can-fail on the wiring: renaming the helper away makes the run print
  `SKIPPED (wiring) — … 72 component(s) were in scope and NONE had their
  NcSelect elements inspected`, not PASS.

### DIFF
--- a/hydra-gates/scripts/lib/check_nc_select_labels.py
+++ b/hydra-gates/scripts/lib/check_nc_select_labels.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: EUPL-1.2
+"""Gate-12 nc-input-labels — every `<NcSelect>` must name itself.
+
+A manual `<label>` next to an `NcSelect` does not associate with the
+component's internal combobox input, so the accessible name has to come
+from the component's own API: `input-label` (rendered as the visible
+label) or `aria-label-combobox`. WCAG 2.2 AA SC 1.3.1 / 4.1.2, ADR-004
+hard rule. Observed 2026-04-30 on doriath.
+
+WHY THIS MOVED OUT OF THE BASH GATE
+-----------------------------------
+The gate used to flatten the file's newlines and run::
+
+    grep -oE '<NcSelect[^>]*>'
+
+`[^>]*` ends the element at the FIRST `>` character in the source. In an
+`NcSelect` that is almost never the end of the tag, because the idiomatic
+vue-select usage puts an arrow function in an attribute value::
+
+    <NcSelect
+        v-model="selectedRoundId"
+        :options="roundOptions"
+        :reduce="(o) => o.id"
+        :input-label="t('scholiq', 'Round')"
+        :aria-label-combobox="t('scholiq', 'Round')" />
+                  ^
+                  the `>` of `=>` — extraction stops here
+
+Everything after `:reduce` is invisible to the gate, so the very props it
+is looking for are cut off and the element is reported as unnamed.
+Measured on scholiq 2026-08-08: **18 findings, 18 of them false** — every
+flagged `NcSelect` already carried `:input-label` AND
+`:aria-label-combobox`, in every case written after the `:reduce` prop.
+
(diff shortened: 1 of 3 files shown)


rubenvdlinde:
Superseded by #269, which landed on `main` as `cdfbd7a`.

This branch no longer merged cleanly (`run-hydra-gates.sh` moved under it), so rather than redo the work, **`check_nc_select_labels.py` and all 17 of its tests were carried into #269 verbatim** and rebased onto current `main`. Two things were added on top:

* the helper now reads `source_scope.markup_mask`, shared with gates 31 and 32, so an `<NcSelect>` named in a `<script>` docblock is not scanned either (#220 / #235 are the same defect in a neighbouring gate);
* a nested `<template #default>` slot no longer ends the SFC template early — the lazy `(.*?)` in the first cut of that mask deleted a real finding at openconnector `EditMapping.vue:376`, caught by measurement before landing.

Measured post-merge on openconnector, attribution-isolated between `main@32fa23e` and `main@cdfbd7a`, from the runner's stdout: gate-12 **13 → 10**, with all three removals verified by hand as elements carrying `input-label=` written after `:selectable="(option) => …"`.

Closing as superseded, not rejected — the analysis in this PR was correct and is quoted in the merged one. The branch is left in place.

Issue #238, opened 2026-08-08
gate-53 effective-manifest-crossref never joins src/registry.js — a manifest component ref and an orphaned registry entry are BOTH silent


rubenvdlinde:
Filed alongside ConductionNL/larpingapp#286, where a registered-but-unreferenced component shipped as unreachable UI long enough for its openspec task to be ticked over it. gates 22 and 53 are the two that exist to catch manifest cross-reference defects, and neither sees this one — in either direction.

## Measured — with ajv resolved

Important, because this gate has a **wiring** failure mode that masks its verdict entirely. Without `NODE_PATH` set, on `ConductionNL/larpingapp@development`:

```
[gate-22] manifest-validation: FAIL — SCHEMA VALIDATION DID NOT HAPPEN — Ajv is not resolvable ...
[gate-53] effective-manifest-crossref: FAIL — SCHEMA VALIDATION DID NOT HAPPEN — the vendored
          validator degraded to its structural lint (Ajv unresolvable mid-run)
```

With `NODE_PATH=<app>/node_modules`, same commit, both go **PASS**. Everything below is from the ajv-resolved runs — a red-for-wiring baseline would have told me nothing.

## Direction 1 — orphaned registry entry (the #286 shape)

`src/registry.js` exports `EventRoster: { kind: 'section', component: EventRoster }`. No `tabs[].component`, no `slots.*`, no `sections[].component` anywhere in `src/manifest.json` names it. The component is resolvable and unreachable; the check-in surface it renders has no entry point.

```
[gate-53] effective-manifest-crossref: PASS
```

## Direction 2 — manifest names a component that does not exist

Control on top of the #286 fix branch: changed one tab to name a component no registry exports.

```json
{ "id": "checkin", "label": "Check-in", "icon": "AccountCheck",
  "component": "ThisComponentDoesNotExistAnywhere" }
```

At runtime `CnObjectSidebar.resolveTabComponent()` logs `Tab "checkin" component "..." not found in registry or customComponents` and renders **nothing** — a blank tab.

```
[gate-22] manifest-validation: PASS
[gate-53] effective-manifest-crossref: PASS
```

Both silent.

## Cause

`scripts/lib/check_manifest_crossref.js` never reads `src/registry.js`. Its joins are manifest→manifest (menu-route → page-id, open-page targets, deepLink correspondence, ADR-044 removals) and manifest→register-JSON (register/schema slug resolution). The one place the component registry comes up, it is explicitly declined:

```js
[...]

Issue #233, opened 2026-08-08
gate-60 icon-vocabulary reports 43 confident FAILs when node_modules is absent — 'Calendar does not exist in vue-material-design-icons'


rubenvdlinde:
When `vue-material-design-icons` is not resolvable, gate-60 does not skip and does not warn — it emits a definite, per-icon verdict claiming the icon **does not exist**. On larpingapp that was 43 findings naming `Calendar`, `Star`, `Trophy`, `Earth`, `MapMarker`, `Package`, `School`, `Sitemap`, `Lightbulb`, `Briefcase` — all of which obviously do exist.

This is a **false RED**, which is rarer and in some ways worse than a false green: it is confidently worded, it names a specific remediation, and following it would replace correct icons with different ones.

## Measured — same tree, same commit, two verdicts

`ConductionNL/larpingapp@development` `c73df92d`, hydra-gates `main` @ `756fe89`:

| run | `node_modules` | gate-60 |
|---|---|---|
| 1 | absent (`npm ci` still running) | `FAIL — 43 icon(s) outside the canonical vocabulary (ADR-077)` |
| 2 | present | `PASS` |

Nothing else changed between the two runs. Sample from run 1's log:

```
FAIL  src/manifest.json: Events: icon "Calendar" does not exist in vue-material-design-icons and is not a vocabulary icon — it renders blank wherever it is not aliased locally.
FAIL  src/manifest.json: XP awarded: icon "Trophy" does not exist in vue-material-design-icons and is not a vocabulary icon — it renders blank wherever it is not aliased locally.
FAIL  src/manifest.json: Setting: icon "Earth" does not exist in vue-material-design-icons and is not a vocabulary icon — it renders blank wherever it is not aliased locally.
...
checked 9 manifest(s): 43 failure(s), 0 warning(s)
```

Positive control: `node_modules/vue-material-design-icons/Calendar.vue` exists in run 2, and the identical manifest then passes.

## Why this matters more than a normal FP

Gates 22 and 53 in the *same run* handled their missing dependency correctly and loudly:

```
[gate-22] manifest-validation: FAIL — SCHEMA VALIDATION DID NOT HAPPEN — Ajv is not
resolvable ... A weaker check reported as a pass is not a pass. Run `npm ci` ...
[gate-53] effective-manifest-crossref: FAIL — ajv not resolvable from ... — refusing
to run fail-open
```

[...]

Issue #232, opened 2026-08-08
gate-38 skip-link treats templates/settings/*.php as a document root — an ISettings fragment cannot carry a bypass link


rubenvdlinde:
gate-38 scopes `templates/settings/*.php` as if each were an app entry point. Those files are `ISettings::getForm()` **fragments** embedded inside core's settings page, which already renders the skip links. Requiring one of them to carry its own is not just unnecessary — satisfying it would be a WCAG regression.

## Measured

Full-tree run against `ConductionNL/larpingapp@development` `c73df92d`:

```
[gate-38] skip-link: FAIL — 1 root component(s) without skip-link / <NcContent>
templates/settings/admin.php: no <NcContent> shell, no skip-link affordance
```

The whole file:

```php
<?php
use OCP\Util;
$appId = OCA\LarpingApp\AppInfo\Application::APP_ID;
Util::addScript($appId, $appId . '-shared-vendor');
Util::addScript($appId, $appId . '-shared-nc-vue');
Util::addScript($appId, $appId . '-settings');
Util::addStyle($appId, 'main');
?>
<div id="settings"></div>
```

## Call path

`lib/Settings/LarpingAppAdmin.php`:

```php
public function getForm(): TemplateResponse
{
    ...
    return new TemplateResponse( ..., 'settings/admin', ... );
}
```

registered via `appinfo/info.xml`:

```xml
<settings>
  <admin>OCA\LarpingApp\Settings\LarpingAppAdmin</admin>
  ...
</settings>
```

Nextcloud renders that response **inside** `/settings/admin`, whose layout is core's `core/templates/layout.user.php`:

```php
<?php if ($_['id-app-content'] !== null) { ?><a href="<?php p($_['id-app-content']); ?>" class="button primary skip-navigation skip-content"><?php p($l->t('Skip to main content')); ?></a><?php } ?>
<?php if ($_['id-app-navigation'] !== null) { ?><a href="<?php p($_['id-app-navigation']); ?>" class="button primary skip-navigation"><?php p($l->t('Skip to navigation of app')); ?></a><?php } ?>
```

So the bypass links are present on the rendered page, one layer above where the gate can grep — the same "I cannot see it is not it is absent" situation the gate **already** carves out for `<CnAppRoot>` two lines earlier in its own source:

```bash
# All 18 fleet apps root on CnAppRoot, so this gate reported every one
# of them as shipping no skip link. ... "I cannot see it" is not
# "it is absent", and only the first of those is true here.
[...]

Issue #230, opened 2026-08-08
gate-58 e2e-networkidle matches COMMENT lines — larpingapp reports 1 finding with zero live calls


rubenvdlinde:
Same shape as gate-48 (#191): the checker greps for the pattern with no comment filter, so a comment *warning against* `networkidle` is reported as a use of it.

## Measured

Full-tree run of `run-hydra-gates.sh` (hydra-gates `main` @ `756fe89`) against `ConductionNL/larpingapp@development` `c73df92d`:

```
[gate-58] e2e-networkidle: FAIL — 1 networkidle wait(s) in changed e2e file(s)
```

The single log line is:

```
tests/e2e/spec-coverage/detail-forms-admin.spec.ts:187:	// live `waitForLoadState('networkidle')` in the suite; every other mention
```

In context (lines 185-189):

```ts
// ADR-074 rule 4: `networkidle` never settles on Nextcloud — the
// notification poll keeps the network permanently busy. This was the LAST
// live `waitForLoadState('networkidle')` in the suite; every other mention
// is a comment warning against it.
```

The repo has **zero** live calls. Every one of the 10 occurrences under `tests/` is a comment:

```
$ grep -rn "networkidle" tests/ src/ | grep -vE "^\S+:[0-9]+:\s*(//|\*|/\*|#)"
$ echo $?
1        # no matches
```

## Cause

`run-hydra-gates.sh` around line 4515:

```bash
grep -nE "waitForLoadState\([[:space:]]*['\"]networkidle['\"]|waitUntil:[[:space:]]*['\"]networkidle['\"]" "$f" \
    | grep -v "e2e-networkidle exclude" \
    | while IFS= read -r hit; do ...
```

The only filter is the `exclude` marker. There is no comment stripping, so `// … waitForLoadState('networkidle') …` matches exactly like a call.

Note this is a **doubly** perverse failure mode: the more carefully a repo documents why it removed its `networkidle` waits, the more findings it accrues. larpingapp is the repo the gate was written against (its `index-pages.spec.ts` is cited in the gate header), so it has the most such comments in the fleet.

## Suggested fix

Strip line comments before matching — drop lines whose match position sits after a `//`, `*`, or `#` on the same line. A minimal version that would fix this case without touching the pattern:

```bash
grep -nE "<pattern>" "$f" \
    | grep -v "e2e-networkidle exclude" \
    | grep -vE "^[0-9]+:[[:space:]]*(//|\*|/\*)" \
```

[...]

PR #241, opened 2026-08-08
fix(gate-54): a relation nested in an array-of-objects is still a property

### LINKED ISSUE #231: gate-54 relation-dialect: property_ids never includes items.properties.* — a correct nested relation is reported as misplaced
Rule-6 check (c) reports a **correctly shaped** relation property as "placed off a property" whenever it is nested inside an array-of-objects.

## Measured

Full-tree run against `ConductionNL/larpingapp@development`:

```
lib/Settings/larpingapp_register.json: x-relation-filter is placed off a property
(inside items / an x-* block / non-property node) — it rides only on the relation
property itself (ADR-062 rule 6)
```

The flagged node is `character.skillOverrides.items.properties.skill`:

```json
{
  "type": "string",
  "format": "uuid",
  "$ref": "skill",
  "x-relation-filter": { "setting": "@object.setting" },
  "description": "The assigned skill whose requirements are waived",
  "title": "Skill"
}
```

That is the canonical dialect exactly — `type` + `format: uuid` + `$ref` + the filter riding on the same property. It is byte-for-byte the same shape as `character.skills` twenty lines above, which the gate accepts. The only difference is nesting depth.

## Cause — call path

`scripts/lib/check_relation_dialect.py`:

1. `property_ids` is populated in a **single, non-recursive** loop (~line 372-388):

```python
property_ids = set()
for sname, schema in schemas.items():
    props = schema.get("properties")
    if not isinstance(props, dict):
        continue
    ...
    for pname, prop in props.items():
        ...
        property_ids.add(id(prop))
```

It only ever walks `schema.properties.*`. Nothing descends into `properties.<p>.items.properties.*`.

2. `_raw_walk()` (~line 470) recurses into **everything**:

```python
def _raw_walk(node, path, property_ids, findings):
    if isinstance(node, dict):
        ...
        if "x-relation-filter" in node:
            if id(node) not in property_ids:
                findings.append((path, "... placed off a property (inside items / an x-* block / non-property node) ..."))
```

So any nested relation property is structurally unrepresentable to check (c): `_raw_walk` finds it, `property_ids` can never contain it, and the finding is unconditional.

## Why this one bites

The finding is **unfixable in the app**. The three ways to clear it are all wrong:

- move the filter off the property — that is the actual rule-6 violation;
- flatten `skillOverrides` from an array-of-objects to something else — a schema redesign to satisfy a linter;
- delete the filter — loses the setting-scoped picker.

So it sits as permanent red. I have left it unfixed in ConductionNL/larpingapp#289 and documented it there rather than mangling correct schema.

## Suggested fix

Collect `property_ids` recursively through `items` (and `items.properties`), mirroring what `_raw_walk` already traverses. Roughly: after adding `id(prop)`, if `prop.get("items")` is a dict with a `properties` dict, add each of those too — and recurse, since arrays of objects can nest.

The other checks in the helper ((b) relation-shape, (f) `$ref` resolution) have the same blind spot for nested properties, so a nested relation missing its `$ref` is currently **not** reported — the gate is simultaneously false-positive on shape and false-negative on the thing that matters. Worth fixing both with one recursive collector.

rubenvdlinde:
Closes #231.

gate-54 check (c) reported a **correctly shaped** relation as *"placed off a property"* whenever it sat inside an array of objects. The finding was **unfixable in the app** — every way to clear it damages correct schema — so it sat as permanent red on any repo with that (common) shape.

## The finding

`ConductionNL/larpingapp@development`, `character.skillOverrides.items.properties.skill`:

```json
{ "type": "string", "format": "uuid", "$ref": "skill",
  "x-relation-filter": { "setting": "@object.setting" }, "title": "Skill" }
```

Byte-for-byte the shape the gate **accepts** twenty lines above on `character.skills`. The only difference is nesting depth.

## Cause

Check (c) works by identity. `_raw_walk()` flags every `x-relation-filter` whose containing dict is not in `property_ids`, and `property_ids` was built in a single **non-recursive** pass:

```python
for sname, schema in schemas.items():
    props = schema.get("properties")
    ...
    for pname, prop in props.items():
        property_ids.add(id(prop))
```

Nothing descended into `properties.<p>.items.properties.*`, so a nested property could never be in the set and the finding was unconditional.

## Fix

`_collect_nested_property_ids()` registers property dicts below a top-level property, through `items.properties` and inline `properties`, bounded at depth 8. A register file is JSON so it cannot contain a cycle; the realistic hazard is pathological depth, and past the bound the walk **over-reports rather than under-reports** — the safe direction.

## New test suite

`scripts/lib/test_check_relation_dialect.py`. The helper had **no** suite, so `tests/run-helper-suites.sh` (which globs `scripts/lib/test_*`) will now execute one.

Every accepted case is paired with an arm that must still be **reported**, so the suite cannot pass against a checker that returns nothing — the failure mode a checker test most easily falls into:

| # | case | expectation |
|---|---|---|
| 1 | top-level relation | accepted (harness positive control) |
| 2 | relation in `items.properties` | accepted — **the regression** |
| 3 | filter on a non-property `items` node | **still reported** |
| 4 | filter in a schema-level `x-*` block | **still reported** |
| 5 | relation two array levels deep | accepted |
| 6 | pathological depth | terminates, over-reports |

```
test_check_relation_dialect: 6 passed, 0 failed
```

**Can-fail proof.** Removing the `_collect_nested_property_ids()` call:

```
test_check_relation_dialect: 4 passed, 2 failed
  FAIL relation inside items.properties is accepted (gate-54 nested FP)
       — got [... x-relation-filter is placed off a property (inside items / an x-* block / non-property node) ...]
  FAIL relation nested two array levels deep is accepted
       — got [... same text ...]
```

Restoring it returns 6 passed, 0 failed.

## Measured on real trees, not just fixtures

larpingapp full-tree gate-54: **4 findings → 3**, and its own `lib/Settings/larpingapp_register.json` goes **clean**. The three that remain are genuine rule-7 findings in `register.d/` fragments, two of which are fixed in ConductionNL/larpingapp#289.

## One thing worth knowing about rule 7 while you are here

Applying rule 7's remediation (add the canonical `$ref`) to `character.ownerRef` in larpingapp **broke the character create dialog**: `character.ocName` already declares `$ref: "player"`, and with a second player-targeted relation on the same schema the dialog stopped rendering a player selector entirely. Reproduced on two independent CI runs (`detail-forms-admin.spec.ts:570`, 1 failed / 170 passed) and green on `development`. Not addressed here — flagging it because rule 7 currently reads as unconditionally safe and is not.

Also noted while measuring, filed separately: #230 (gate-58 matches comment lines), #232 (gate-38 scopes `ISettings` fragments), #233 (gate-60 reports 43 confident FAILs when `node_modules` is absent), #238 (gate-53 never joins `src/registry.js`).

### DIFF
--- a/hydra-gates/scripts/lib/check_relation_dialect.py
+++ b/hydra-gates/scripts/lib/check_relation_dialect.py
@@ -337,6 +337,46 @@ def _has_uuid_format(prop):
     return False
 
 
+def _collect_nested_property_ids(prop, out, _depth=0):
+    """Register property dicts nested BELOW a top-level schema property.
+
+    A relation can legitimately live inside an array of objects — e.g.
+    `character.skillOverrides.items.properties.skill`, which carries
+    type/format/$ref and its `x-relation-filter` on the property itself,
+    byte-for-byte the same shape the gate accepts one level up.
+
+    Check (c) works by identity: `_raw_walk()` flags every `x-relation-filter`
+    whose containing dict is not in `property_ids`. That set used to be built
+    from `schema["properties"]` only — a single, non-recursive pass — so a
+    correctly-shaped nested relation was structurally unrepresentable to the
+    check and reported as "placed off a property" unconditionally. The finding
+    was unfixable in the app: the three ways to clear it (move the filter off
+    the property, flatten the array-of-objects, or drop the filter) are all
+    worse than the code being flagged.
+
+    Observed 2026-08-08 on larpingapp (ConductionNL/.github#231).
+
+    `_depth` bounds the walk; schemas nest a couple of levels at most, and an
+    unbounded recursion here would be a denial-of-service on a malformed doc.
+    """
+    if _depth > 8 or not isinstance(prop, dict):
+        return
+
+    # Arrays of objects: items.properties.*
+    items = prop.get("items")
+    if isinstance(items, dict):
+        _collect_nested_property_ids(items, out, _depth + 1)
+
+    # Inline object properties: properties.*
+    nested = prop.get("properties")
+    if isinstance(nested, dict):
+        for nname, nprop in nested.items():
+            if not isinstance(nprop, dict) or nname.startswith("@"):
+                continue
+            out.add(id(nprop))
+            _collect_nested_property_ids(nprop, out, _depth + 1)
+
+
 def _resolve_ref(ref, keys):
     """Return ('ok'|'warn'|'fail', normalized). Numeric → warn (live-schema
     form). Hash form resolves to its last path segment."""
@@ -386,6 +426,8 @@ def check_file(path, keys, findings, base_ref):
             if not isinstance(prop, dict) or pname.startswith("@"):
                 continue
             property_ids.add(id(prop))
+            # A relation nested in an array-of-objects is still a property.
+            _collect_nested_property_ids(prop, property_ids)
             pline = plines.get(pname, 0)
             in_diff = changed is None or pline in changed
 
--- a/hydra-gates/scripts/lib/test_check_relation_dialect.py
+++ b/hydra-gates/scripts/lib/test_check_relation_dialect.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+"""test_check_relation_dialect.py — suite for check_relation_dialect.py (gate-54).
+
+Picked up automatically by tests/run-helper-suites.sh (it globs
+scripts/lib/test_*), so it runs in CI the moment it lands.
+
+The suite is written so that EVERY assertion has a paired arm that must go the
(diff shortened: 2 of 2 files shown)


rubenvdlinde:
Superseded by #255.

This branch fixed only the false-positive half: it added nested property ids to `property_ids` but never iterated them, so checks (b) relation-shape, (d) filter tokens and (f) $ref resolution stayed blind to nested properties — the false negative #231 explicitly asks to close stayed open. Its test file also contained no runnable `def test_`.

#255 uses one recursive collector that feeds all the checks, keeps this branch's depth-cap fail-safe assertion (carried over as a test), and ships 24 mutation-checked tests.

Issue #242, opened 2026-08-08
Three gates report PASS over an EMPTY DIFF without calling _skip — so --require-full-coverage cannot see them (openconnector: 404 + 32 hidden findings)


rubenvdlinde:
## Summary

`--require-full-coverage` is built on `_skip <n> <name> <category>`: a gate that inspected nothing must declare itself `structural` or `wiring`, and those count against coverage. **Three gates bypass that entirely.** They are unconditionally diff-scoped, and when the diff is empty they call `_pass`, not `_skip`.

The result is a coverage assertion that reports complete while three gates inspected zero files, and there is no line anywhere in the output that says so.

Measured on `ConductionNL/openconnector@development` (f2028a96) with the gate package cloned fresh from `.github@main` (756fe89).

## The three

| gate | helper | line | what it prints |
|---|---|---|---|
| 19 e2e-coverage | `scripts/lib/check_e2e_coverage.py` | 728 | `PASS — no spec files in diff` |
| 25 contract-coverage | `scripts/lib/check_contract_coverage.py` | 353 | `PASS — no new public endpoints in diff` |
| 63 settings-surface | `scripts/lib/check_store_and_settings_surface.py` | 277 | `No changed manifest / menu-layout — gate skipped (ADR-020 diff scoping).` |

gate-63 is the clearest: **its own log says "gate skipped" while its verdict line says PASS.** Two statements about one measurement, on the same run.

## Why the diff is empty

`check_e2e_coverage.py:724`:

```python
base_ref = os.environ.get("HYDRA_GATE_BASE_REF", "origin/development")
touched = changed_spec_files(base_ref, app_dir)
if not touched:
    print(f"[gate-{GATE_NUM}] e2e-coverage: PASS — no spec files in diff")
    return 0
```

This reads `HYDRA_GATE_BASE_REF` **regardless of whether the runner was given `--scope-to-diff`**. So on any run whose HEAD is `origin/development` — a push to the mainline, the `development → beta` sync PR, or anyone measuring the tree as it actually stands — the diff is empty *by construction* and the gate passes having read nothing.

This is the same base==HEAD shape as .github#183/#179, but it survives those fixes because it lives in the helpers, below the runner's base resolution.

## What it hides, measured

Running the helpers directly with the base set to the repository's root commit (`ea8838bd`, i.e. every file is "changed"):

```
[gate-19] e2e-coverage:      FAIL — 404 scenario(s) without a running e2e test
[...]

PR #249, opened 2026-08-08
fix(gate-19): read the test file with a parser, not three regexes (#234, #239, #244)

### LINKED ISSUE #234: gate-19: a TRAILING COMMA before the closing paren makes a real test read as an empty body — false RED
## Summary

`check_e2e_coverage.py`'s `_enclosing_block()` finds a test's body by scanning **back from the closing paren** and requiring the first non-whitespace character to be `}`:

```python
j = len(whole) - 2                      # skip the final ')'
while j >= 0 and whole[j].isspace():
    j -= 1
body = ""
if j >= 0 and whole[j] == "}":
    ...
```

A **trailing comma** after the callback puts a `,` there instead. `body` stays `""`, the empty-body rule fires, and a real, asserting test is reported as *"referenced only by a test that never runs"*.

This is a **false RED**: it sends someone to fix a test that already works, or to add an `@e2e exclude` to a scenario that is genuinely covered.

## Minimal reproduction

```ts
// @e2e s::x
test(
  'name',
  async ({ page }) => {
    await expect(page).toBeTruthy()
  },
)
```

Against `origin/main` **and** against the current tip (this bug is untouched by #212):

```
trailing-comma test         live = False   <- WRONG
no-trailing-comma control   live = True
```

The only difference between the two inputs is the `,` before `)`.

## Measured in the fleet

`softwarecatalog` `tests/e2e/org-archimate-export.spec.ts:303` —
`org-archimate-export::user-triggers-organization-export-with-toggles`. The declaration ends:

```
...toBe('true')
		}
	},
	)
```

`_enclosing_block()` returns `bodylen=0` for it. The test drives a real combobox, real checkbox toggles and a real button click, and asserts the outgoing request shape.

It stopped being reported by accident in #212: once `test.describe(` became visible to the declaration regex, the tag began resolving outward to the enclosing `test.describe`, whose body computation succeeds. **The underlying bug is unchanged** — any tag that still resolves directly to a trailing-comma `test(...)` is still wrongly dead.

## Suggested fix

Skip a single trailing `,` (and any whitespace around it) before the `}` test in `_enclosing_block()`. Prettier and ESLint's `comma-dangle: always-multiline` both produce this shape, so it is the default formatting for a multi-line call in most of the fleet, not an oddity.

## Both-ways proof it will need

- a trailing-comma test with a real body -> **live** (the fix)
- a trailing-comma test with an **empty** body -> **dead** (the control — the empty-body rule must survive)
- a trailing-comma `test.skip(...)` -> **dead**
- the existing no-trailing-comma cases -> unchanged

## Not doing it in #212

#212 fixed two specific defects (#192, #210) and each of its changes is measured against the fleet. Folding in a third behaviour change — one that moves refs from dead to live for an unrelated reason — would have made that PR's fleet numbers unattributable. Filed separately on purpose.

### LINKED ISSUE #239: gate-19: a CONDITIONAL `test.skip(true, reason)` inside an `if` guard is read as a permanent skip — a test that ran and PASSED is reported as "never runs"
## Summary

`check_e2e_coverage.py`'s "does this test actually run" discriminator classifies **any** `test.skip(true, …)` in a test body as an unconditional skip. But `test.skip(true, reason)` is also the correct Playwright form for a *conditional* skip written inside an `if` guard — the single most common defensive idiom in the fleet's e2e suites. The `true` is the API's "skip from this point" shape; the **call site** is what carries the condition.

Result: gate-19 reports "referenced only by a test that never runs" about tests that run on every reachable environment — including, provably, tests that **passed in the same CI run that produced the finding**.

## Minimal reproduction

Two tests in one file, identical but for a guarded skip:

```ts
// @e2e openspec/specs/demo/spec.md#s1-guarded-skip-inside-an-if
test('guarded: runs whenever the app is reachable', async ({ page }) => {
	const response = await page.goto('/app').catch(() => null)
	if (!response) {
		test.skip(true, 'app not reachable')
		return
	}
	await expect(page.locator('body')).not.toContainText('Internal Server Error')
})

// @e2e openspec/specs/demo/spec.md#s2-no-skip-at-all
test('clean: no skip anywhere', async ({ page }) => {
	await page.goto('/app')
	await expect(page.locator('body')).not.toContainText('Internal Server Error')
})
```

```
$ HYDRA_GATE_BASE_REF=base python3 check_e2e_coverage.py .
demo::s1-guarded-skip-inside-an-if — @e2e tag present but the test does not run:
  referenced only by a test that never runs (tests/e2e/demo.spec.ts). …
[gate-19] e2e-coverage: FAIL — 1 scenario(s) without a running e2e test
```

S2 passes, S1 fails. The only difference is the guarded skip.

## Live confirmation

On ConductionNL/procest PR #765, gate-19 flagged `req-zak-004a` and `req-zak-004b`. Both are tagged on:

```ts
test('cases index renders so a case dossier tab can be opened', async ({ page }) => {
	const response = await page.goto('/index.php/apps/procest/cases').catch(() => null)
	if (!response) {
		test.skip(true, 'Procest dev container not reachable')
		return
	}
	await expect(page.locator('body')).not.toContainText('Internal Server Error', { timeout: 10000 })
})
```

No `.skip` / `.fixme` modifier. Its only skip is inside `if (!response)`. The E2E job on the same commit reported **87 passed / 0 failed / 38 skipped** — this test is one of the 87. gate-19 said it never runs.

## Why the current rule was chosen, and why the reasoning inverts here

The source comment is explicit and deliberate (around L440–L460):

> `test.skip(true)` — an UNCONDITIONAL skip at the top of a body … so the discriminator is the argument, not the call.

That holds for `test.skip(true)` written *at the top of a body*, which is a permanent skip. It does not hold for the same call written *inside a conditional block*, where the guard is the condition and `true` is only how you tell Playwright "skip now". Both forms are legal, idiomatic Playwright and are textually identical at the call.

## Blast radius

Counted with the gate's own patterns across the accessible fleet checkouts — occurrences of `test.skip(true, …)` that are **inside** an `if` / `else` / `catch` guard:

| repo | guarded `test.skip(true, …)` |
|---|---|
| openregister | 81 |
| procest | 22 |
| hermiq | 4 |
| docudesk | 3 |
| nldesign | 1 |

**111 occurrences**, versus **4** genuinely unconditional ones in the same checkouts (all in one procest file, now fixed). So the rule currently misfires on ~96% of the call sites it inspects, and any PR touching one of those specs gets a false finding.

## Why this is worse than a normal false positive

The remediation gate-19 offers is *"replace the tag with a reason-bearing `@e2e exclude`"* — i.e. declare the scenario **not** e2e-covered. Complying with a false finding therefore **deletes a true coverage claim** and permanently marks a genuinely-tested scenario as untestable. The gate actively pushes the codebase in the wrong direction.

## Suggested fix

Determine conditionality from the **call site**, not the argument: a `test.skip(true, …)` is unconditional only when it is a direct statement of the test/describe body, not nested inside an `if` / `else if` / `else` / `catch` / `try` block. The existing `_has_own_unconditional_skip()` already walks block structure to exclude skips belonging to nested *tests* — the same walk can note enclosing conditional blocks.

Include the reproduction above as a fixture: it fails both ways today (S1 false-positive; and a genuine top-of-body `test.skip(true)` must still be caught).

## Secondary observation — no representation for "quarantined pending an issue"

The same procest PR converted three permanently-skipped tests to `test.fixme` referencing a tracking issue. gate-19 correctly notes they do not run, but its only sanctioned end state is `@e2e exclude <reason>`, which means "this scenario is out of scope for e2e, forever". A scenario that *should* be e2e-tested and is blocked on a tracked fixture gap has no way to say so. Consider accepting a tracked form (e.g. `@e2e blocked <issue> <reason>`) that is visible in reporting without being either a silent pass or a permanent exclusion.

Found during a fleet-wide triage of disabled e2e declarations.

### LINKED ISSUE #244: gate-19: 'tag present but the test does not run' fires on tests that DO run — 34 of 190 findings on nldesign (18% inflation), disproved by the same CI run
## Summary

Measuring nldesign's true gate-19 debt (root-commit-scoped, per #242) gives **190 findings**. **34 of them are false positives** in one specific branch: `@e2e tag present but the test does not run: referenced only by a test that never runs (<file>)`.

Every file named in those 34 has **passing tests in the same CI run**. The real debt is **156**.

This is the counting hazard in its concrete form: a finding count is not a defect count, and here the inflation is 18%.

## Measured

```
$ run-hydra-gates.sh --scope-to-diff --base <root commit>
[hydra-gates] SCOPE-FILE-COUNT: 3611
[gate-19] e2e-coverage: FAIL — 190 scenario(s) missing @e2e
```

Split by reason (191 log lines, 191 distinct — no duplicate-emission inflation, so the 34 are the whole story):

| reason | count | verdict |
|---|---|---|
| `— missing @e2e` | **156** | genuine |
| `— @e2e tag present but the test does not run` | **34** | **false positive** |

Files named in the 34, against passing tests in nldesign CI job `93093896316`:

| file | flagged as "never runs" | tests that PASSED in that run |
|---|---|---|
| `admin-settings.spec.ts` | 14 | **17** |
| `token-editor-ui.spec.ts` | 6 | **8** |
| `app-theming.spec.ts` | 5 | **4** |
| `token-set-apply-dialog.spec.ts` | 3 | **5** |
| `token-set-contrast-audit.spec.ts` | 2 | **2** |
| `custom-token-set-upload.spec.ts` | 2 | **4** |
| `token-set-apply-toggle-behavior.spec.ts` | 1 | **3** |
| `token-import-export.spec.ts` | 1 | **2** |

Not one of them is a file whose tests never run.

## Five flagged anchors matched to passing tests, verbatim

The scenario names and the CI `✓` lines are the same tests:

```
admin-settings::settings-panel-appears-in-admin-area                    -> ✓ 10 admin-settings.spec.ts:24:6  › Settings panel appears in admin area (3.2s)
admin-settings::settings-panel-position-relative-to-nextcloud-theming   -> ✓ 11 admin-settings.spec.ts:35:6  › Settings panel position relative to Nextcloud theming (2.6s)
admin-settings::dropdown-populated-with-token-sets                      -> ✓ 12 admin-settings.spec.ts:78:6  › Dropdown populated with token sets (2.6s)
admin-settings::dropdown-label-is-associated-with-select                -> ✓ 13 admin-settings.spec.ts:95:6  › Dropdown label is associated with select (2.6s)
admin-settings::preview-box-renders-with-token-set-colors               -> ✓ 15 admin-settings.spec.ts:133:6 › Preview box renders with token set colors (2.8s)
```

`admin-settings.spec.ts` contains `test.describe('admin-settings', () => {` — a plain describe, **not** `.skip` — 17 `test(` blocks, and **zero** `test.skip` / `test.fixme` / `describe.skip` anywhere in the file.

## Likely mechanism

nldesign writes the tag **inside** the `test(` call, between the opening paren and the title:

```ts
test(
    // @e2e openspec/specs/admin-settings/spec.md#settings-panel-appears-in-admin-area
    'Settings panel appears in admin area',
    async ({ page }) => {
        await page.goto(THEMING_URL)
        ...
    },
)
```

The helper's forward search from the tag looks for the `test(` that owns it — but here `test(` is on the line **above** the tag. The search runs forward, so it either finds the *next* test's declaration or runs off the end of the file, and the "no live declaration found" path reports *"referenced only by a test that never runs"*.

That fits the distribution: **14 of 17** tagged tests in `admin-settings.spec.ts` are flagged, and the file's tag layout is uniform.

The helper already carries a comment about this class of bug — *"never runs", about a file whose tests run fine* — so the ancestor/`describe.skip` direction has been fixed. The declaration-**above**-the-tag layout looks like the remaining one.

## Suggested fix

When resolving a tag to its owning test, search **backwards** as well as forwards for the nearest `test(` / `it(` declaration, and prefer the nearer of the two. Both layouts are in fleet use:

```ts
// @e2e ...          |   test(
test('title', ...)   |       // @e2e ...
                     |       'title',
```

Whatever the fix, the discriminator worth adding is: **"I could not resolve this tag to a declaration" is not "the test never runs."** They are different facts and only the second is a finding — the same distinction gate-5 already draws between "the attribute is absent" and "I could not open the class".

## Why this was only visible at the root commit

A normal run scopes to the diff, so these never appear. A `Scope: full repo` run does not help either: gate-19 is unconditionally diff-scoped **inside its helper**, so an empty diff makes it call `_pass` (#242). It takes `--scope-to-diff --base <root>` to see any of it.

Related: #242 (gates 19/25/63 invisible to `--require-full-coverage`), #243 (`set -e` leak aborts the run mid-way).

rubenvdlinde:
Closes #234. Closes #239. Closes #244.

## The three issues were one decision

Gate-19 reads JavaScript. It was reading it with regular expressions and a
hand-rolled paren walk, and all three open false-positive issues came out of
that — reported as the same sentence, *"referenced only by a test that never
runs"*, about tests that ran and **passed in the same CI run**.

| # | cause |
|---|---|
| **#234** | The body was located by stepping back from `)` over whitespace and requiring a `}`. **Prettier's default** and ESLint's `comma-dangle: always-multiline` put a `,` at exactly that index, so `body` stayed `""` and the empty-body rule fired on a real, asserting test. |
| **#239** | The discriminator was the **argument alone**. But `true` is just Playwright's "skip from this point" shape — the **call site** carries the condition. 111 guarded call sites in the fleet against 4 genuinely unconditional ones. |
| **#244** | **Diagnosed, not assumed** — see below. |

### What #244 actually was

The issue guessed *"the search runs forward, so it either finds the next
test's declaration or runs off the end"*. The first half is exactly right and
it is the whole mechanism. nldesign writes every tag **between the open paren
and the title**:

```ts
test(
    // @e2e openspec/specs/admin-settings/spec.md#settings-panel-appears-in-admin-area
    'Settings panel appears in admin area',
    async ({ page }) => { … },
)
```

A forward-only search binds each tag to the **next test in the file**. That
mis-binding then met **#234** on whichever test it landed on — every one of
those declarations ends `},\n)` — so the wrong test *also* read as an empty
body. **Two defects, one symptom**, which is why the fixture asserts the
binding directly and not only the count.

### Why #239 is worse than an ordinary false positive

The remedy the gate prints is *"replace the tag with a reason-bearing
`@e2e exclude`"*. Complying with a false #239 finding therefore **deletes a
true coverage claim** and permanently marks a genuinely-tested scenario as
untestable. The gate was pushing the codebase in the wrong direction.

## The fix: parse, don't grep

Tokenise the file once — comments, string contents, template contents and
regex literals blanked, offsets and newlines preserved so tags found in the
original text locate into the structure. **String delimiters are kept**,
because "is the first argument a string literal" is the whole difference
between `test.skip('title', fn)` (a switched-off declaration) and
`test.skip(cond, 'reason')` (a statement in a running test).

Then build the real tree of `test`/`describe` calls with header and body
ranges, and answer structure questions from it.

Everything the old regexes had earned is kept and re-asserted:
`rx.test(` is not Playwright's `test()`; `latest(`/`submit(` merely end in a
name; **`test.describe.skip(` must match — #212 is NOT undone**; `.only` /
`.serial` / `.parallel` are not switched-off markers; and
`test.beforeEach(` / `test.use(` / `test.step(` / `test.setTimeout(` /
`test.describe.configure(` are not declarations at all.

## Signalling

This gate returned its finding **count** as an exit status — a byte — so 266
findings left as 10, and 256 would have left as **0**, read as PASS (#209).
The clamp that fixed the wrap made the byte carry **neither**: a 404-finding
run exited **255** while stdout said **404** (#242). Two numbers for one
measurement means one came through a lossy channel.

The byte is a status now and nothing else — **0 pass, 1 fail, 2 error** — and
the count is on stdout where the runner already reads it. A crash reports
`SKIPPED (wiring)`, visible to `--require-full-coverage`, instead of a
fabricated verdict; the runner also stops discarding the helper's stderr.

Verified end to end through the real runner: `FAIL — 159 scenario(s)` with
exit 1, and a planted crash yields
`SKIPPED (wiring) — check_e2e_coverage.py exited 2 (error)`.

## Not touched

The empty-diff `_pass` branch. That is **#242**'s subject and is being fixed
separately — I only renamed its literal `0` to `EXIT_PASS`.

## Measured — 24 local checkouts, root-commit-scoped

`--scope-to-diff --base <root-commit>`. **8790 → 8698 findings, −92.** Every
one of the 92 is a false positive removed; **not one finding was added
anywhere.**

| repo | before | after | Δ |
|---|---|---|---|
| nldesign | 190 | **156** | −34 (exactly the 34 in #244) |
| procest | 1181 | 1166 | −15 |
| softwarecatalog | 312 | 291 | −21 |
| decidesk | 991 | 984 | −7 |
| shillinq | 284 | 279 | −5 |
| openregister | 799 | 794 | −5 |
| opencatalogi | 51 | 46 | −5 |
| openconnector | 412 | 412 | 0 — its 6 dead refs are real `test.describe.skip` |
| openbuild | 187 | 187 | 0 — its 36 are real `test.skip('title', …)` |
| larpingapp | 101 | 101 | 0 — its 23 are real `test.fixme` |
| scholiq | 102 | 102 | 0 |

## Planted true positives — the gate still catches real gaps

Against **nldesign's real spec and real e2e suite, after the fix** (baseline
156):

- a scenario with **no test at all** → caught (`missing @e2e`)
- a scenario tagged only by a **skipped** test → caught
- a scenario tagged only by an **empty-bodied** test → caught
- a fourth scenario tagged by a **real** test in nldesign's own
  trailing-comma layout → correctly **not** flagged

**156 → 159.** The fix narrows nothing.

## Tests

**75 → 107**, all green, plus the 27 other helper suites and the 59
entry-point tests.

Mutation-checked — reinstating each defect turns the right tests red:

| mutant | red |
|---|---|
| trailing-comma bug back | 2 |
| argument-only skip rule back | 4 |
| forward-only tag resolution back | 6 |
| header branch deleted | 1 |
| count back in the exit byte | 2 |
| **`_ref_is_live` returns True always** (the anti-widening control) | **25** |

⚠️ One earlier mutant **survived**: deleting the header branch left the whole
suite green, because a `test()` header has no children so the fallback
returned the same node. That fixture could not see the branch it was meant to
cover. A describe-header case was added, and it kills the mutant.

### DIFF
--- a/hydra-gates/scripts/lib/check_e2e_coverage.py
+++ b/hydra-gates/scripts/lib/check_e2e_coverage.py
@@ -79,7 +79,12 @@ Diff scope
 In gate mode (default), the gate is diff-scoped via ``HYDRA_GATE_BASE_REF``
 (default ``origin/development``): only scenarios in spec files that are ADDED or
 MODIFIED in the PR are checked. Scenarios in untouched spec files are never
-flagged. Exit code = number of uncovered (or bare-exclude) scenarios.
+flagged.
+
+Exit code is a STATUS, not a count: ``0`` pass, ``1`` fail, ``2`` error. The
+number of findings is on stdout, in the ``FAIL — <n> scenario(s)`` summary
+line. Read stdout; an exit status is one byte and this gate has already
+returned a count through it twice.
 
 Report mode (``--mode report``) scans the entire ``openspec/specs/`` tree and
 emits a JSON summary — not diff-scoped, always exits 0.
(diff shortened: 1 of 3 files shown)


rubenvdlinde:
## Correction — the test count in this PR's description is wrong

The description and the commit message both say **75 → 107**. The true number is **75 → 105**.

```
$ python3 hydra-gates/scripts/lib/test_check_e2e_coverage.py   # at 7b66766, merged main
Ran 105 tests in 0.911s
OK

$ grep -c '    def test_' hydra-gates/scripts/lib/test_check_e2e_coverage.py
105
```

I wrote 107 from memory of an intermediate run (103) plus the two tests I added afterwards, without re-reading the runner's own line — the exact habit this gate's history is about. Nothing else in the measurements is affected: the suite is green, every other figure in the description came from a captured run, and the fleet numbers were re-verified against the merged `main` below.

Everything else stands: 27 other helper suites pass, 59 entry-point tests pass, and all six mutants are killed.

Issue #268, opened 2026-08-08
hydra-gates: an EMPTY ADR-020 diff scope is classified 'structural', so require-full-coverage fails every PR that touches no spec and no manifest (exit 98)


rubenvdlinde:
## Summary

Since `c1f7deb` (merge of #258, `main` at **2026-08-08T14:21:55Z**), gates **19 / 25 / 62 / 63** classify an **empty ADR-020 diff scope** as `structural`. `structural` counts against `--require-full-coverage`, which is ON by default. The result: **any PR whose diff happens not to touch a spec file and not to touch a manifest/menu-layout fails with exit 98**, with zero gate findings and nothing wrong with the code.

This is not the same as #169 (gates 4/24/33, subject matter genuinely absent from the repo). Here the subject matter is present, clean, and passes when the gate is allowed to look at it.

## Reproduction

Repo: `ConductionNL/doriath`, PR #191 — a 3-file change (one controller + two new unit test files). Every gate that ran passed.

```
##[error]hydra-gates passed every gate that ran, but a gate whose SUBJECT MATTER EXISTS
did not report, and hydra-gates-require-full-coverage is set.
GATES THAT DID NOT RUN: 19 62 63
```

Reproduced locally against `hydra-gates` @ `846baed`:

```
$ bash .../run-hydra-gates.sh --require-full-coverage --scope-to-diff origin/development .
[gate-19] e2e-coverage:     SKIPPED (structural) — the diff touched NO spec file …
[gate-62] store-plane:      SKIPPED (structural) — the diff touched no manifest or menu-layout …
[gate-63] settings-surface: SKIPPED (structural) — the diff touched no manifest or menu-layout …
[hydra-gates] 57 GATE(S) GREEN — but 3 of 60 APPLICABLE gates DID NOT RUN
EXIT=98
```

## The subject matter exists, is inspectable, and is clean

Run gate-63's own checker on the same tree with and without `--base`:

```
$ python3 scripts/lib/check_store_and_settings_surface.py . --gate settings
rc=0
checked 2 manifest(s): 0 failure(s), 0 warning(s).

$ python3 scripts/lib/check_store_and_settings_surface.py . --gate settings --base origin/development
rc=3
No changed manifest / menu-layout — EMPTY SCOPE (ADR-020 diff scoping against 'origin/development').
2 manifest(s) exist here and NONE were inspected; this gate's placement rules are UNVERIFIED by this run.
```

[...]

Issue #252, opened 2026-08-08
gate-19: a test QUARANTINED behind a tracked blocker has no honest end state — its only remedy is to declare the scenario permanently out of e2e scope (59 findings measured)


rubenvdlinde:
Split out of #239's secondary observation, which #249 deliberately did not address because it is a **policy** change rather than a false-positive fix.

## The gap

gate-19 is correct that a `test.fixme` / `test.skip('title', …)` does not run, so the scenario is not proven. But the only remedy it prints is:

> replace the tag with a reason-bearing `@e2e exclude`

and `@e2e exclude` means *"this scenario is out of scope for e2e, forever"*. A scenario that **should** be e2e-tested and is blocked on a tracked fixture or product gap has no way to say so. The honest states are three, and the vocabulary has two.

## Measured, root-commit-scoped, against `.github@main` (7b66766)

These are the findings that survive #249 — i.e. every one is a **correct** report of a test that genuinely does not run:

| repo | findings | what they actually are |
|---|---|---|
| openbuild | 36 | `test.skip('REQ-OBR-002 — builder route mounts a nested CnAppRoot …', …)` — 8+ declarations in `openbuild-runtime.spec.ts` alone |
| larpingapp | 23 | `test.fixme(…)`, every one carrying `// FIXME(detail-actions-blocker): detail Actions menu never renders — DETAIL_ACTIONS_BLOCKER.` |
| procest | 26 | `test.skip(true, 'Requires a seeded case fixture in the dev container')` at top of body |
| openconnector | 6 | `test.describe.skip('Catalog page — browse, filter, badges (REQ-001)', …)` |
| scholiq | 2 | |

**93 findings**, and the 59 in larpingapp + openbuild are the clearest case: the author already wrote down the blocker (`DETAIL_ACTIONS_BLOCKER`, `REQ-OBR-*`) in a comment the gate cannot read. Closing these honestly today means asserting something false.

## What the vocabulary is missing

Three distinct facts, two words:

| fact | today | should be |
|---|---|---|
| not suitable for e2e (backend contract, covered by PHPUnit/Newman) | `@e2e exclude <reason>` | unchanged |
| should be e2e-tested, blocked on a tracked issue | **no representation** — pressure to use `@e2e exclude` | e.g. `@e2e blocked <issue> <reason>` |
| should be e2e-tested, nobody has written it | `missing @e2e` | unchanged |

## Why it matters more than a wording nit

[...]

Issue #191, opened 2026-08-06
gate-48 csrf-cochange matches a REMOVED COMMENT — nldesign is red for a docblock that only names the attribute


rubenvdlinde:
## What happened

`gate-48 csrf-cochange` is red on **nldesign** (`development` vs `origin/beta`, hydra-gates @main). Nothing about CSRF changed in that diff. The gate matched a **removed line of prose inside a class docblock**:

```
- * (#[PublicPage] + #[NoCSRFRequired]) and the response contract are owned by
```

That line is part of a docblock rewrite describing *where the auth posture is owned*. No attribute and no annotation was removed; the surrounding diff replaces one explanatory paragraph with another.

## Why

The detector matches the token anywhere on any removed line:

```bash
_csrf_removed=$(git diff -U0 "${BASE_REF}...HEAD" -- 'lib/Controller/*.php' 2>/dev/null \
    | grep -E '^-.*(@NoCSRFRequired|#\[NoCSRFRequired\])' || true)
```

`^-.*` puts no constraint on code position, so a comment that *names* the attribute is indistinguishable from an attribute that was deleted.

This is the same defect class as #184 (gate-64 grepped a quoted string literal, so it matched comments and missed constants). A checker that greps a token rather than a code position fails in both directions.

## Why this one is especially bad to leave

The fleet is currently being driven to green. The cheapest way to make this finding disappear is to **reword a comment** — which changes nothing about CSRF and teaches exactly the habit the gate exists to prevent. A gate that can be satisfied by editing prose is worse than no gate, because it manufactures the appearance of a security review.

## Suggested fix

Require the token in a **code position** rather than anywhere on the line:

- attribute form: after `-` and optional whitespace the content starts with `#[`, and the attribute list contains `NoCSRFRequired`;
- legacy docblock form: after `-`, optional whitespace and an optional leading `*`, the content starts with `@NoCSRFRequired`.

The nldesign line fails both (its `#[NoCSRFRequired]` sits mid-sentence after `(`), while a genuine deletion of either form still matches.

## Mutation check to include

The fixture needs both arms, or the fix is unfalsifiable:

1. **must stay green** — a diff that only removes a comment line mentioning `#[NoCSRFRequired]` in prose (the nldesign line verbatim).
2. **must go red** — a diff that removes a real `#[NoCSRFRequired]` attribute line with no frontend CSRF signal, and separately one that removes a real ` * @NoCSRFRequired` docblock tag.

[...]

PR #280, opened 2026-08-09
fix(gates 45-55): eleven gates passed over an unopened scope, eight over a dead interpreter, and three could not see the defect they exist for

rubenvdlinde:
fix(gates 45-55): eleven gates passed over an unopened scope, eight over a dead interpreter, and three could not see the defect they exist for

Every gate in this band was given ONE textbook true positive of exactly what
it exists to catch, planted in a real fleet repo, then removed again. Where a
gate could not fail, it was repaired; where it could, the plant is now a
regression test. Measured at package sha 34370f6.

## 1. All eleven reported PASS over a scope they never opened (#242/#240/#258/#268)

On a README-only diff against larpingapp, gates 45-55 printed eleven PASS
lines and the summary read "53 of 53 applicable gates ran". Not one of them
had opened a file. Gates 4/6/7/19/25/28/62/63 have answered the identical
situation with NOT APPLICABLE since #268; this band never adopted it.

Gates 47 and 48 are the sharper case: they can only answer a question about a
CHANGE SET, so on every builder full-repo run in the fleet — no base ref at
all — they printed a co-change verdict they had not formed.

## 2. Eight reported PASS over a crashed interpreter (#147/#249/#262)

A planted defect only fires when the gate runs, so no plant can see this. With
a `python3` on PATH that exits 1 on every call, on a tree carrying real
findings:

  gate-46  PASS — over the 277 unresolved @spec findings, across 104 distinct
           targets, it had reported one run earlier on the same files
  gate-47  PASS — on the same diff where it had just reported FAIL
  gate-45/49/50   PASS  (`2>/dev/null` discarded status and traceback)
  gate-51/54/55   PASS  (`|| true` discarded the status)
  gate-52  FAIL — "1 custom-widget finding(s)", a fabricated finding: the
           helper returned its COUNT as its exit status, the same channel
           Python uses for a traceback (#209). The count was also clamped to
           99 to fit in a byte. It now prints `findings=N` on stdout and exits
           boolean; no `findings=` line means the helper died.

gate-54 was the quietest: its advisory WARN half reads the same log, so a dead
helper silenced both halves at once.

## 3. gate-45 was the residue of #272's fix (.github#274)

#272 migrated gates 35/40/42/44 off `[ -d src ]` onto `_a11y_has_markup_dir`
and left the twelfth member of the family behind. On a templates-only app
gate-45 reported NOT APPLICABLE — "this repo ships no frontend" — over a
`<style>` block with `transition:` and no reduced-motion fallback, in the same
file gate-43 FAILED on in the same run. `na` is the one verdict that removes a
gate from coverage accounting.

The regression test was already written and gate-45 was excluded from it by
name, with a comment explaining why. Removing the name from ARM 4's skip list
in test_gate_a11y_markup_scope.sh IS the test; it fails against 34370f6.

## 4. gate-47: prose satisfied it, and a qualified attribute did not

`_ANNOTATION_RE` was an unanchored alternation of string literals, and it was
wrong in both directions from that one regex — the pairing #269 found in
gate-48 and never carried to its sibling.

  FALSE POSITIVE  rewording ONE docblock sentence that merely NAMES the
                  annotation ("becomes `@NoAdminRequired` again, paired with a
                  real ownership check") made the gate demand a test
                  co-change. A gate satisfiable by prose manufactures the
                  appearance of a security review (#191).
  FALSE NEGATIVE  `#[\OCP\AppFramework\Http\Attribute\NoAdminRequired]` was
                  invisible. A commit adding exactly that to a controller —
                  opening an admin-only endpoint to every authenticated user —
                  with no test in the diff reported PASS.

Now position-anchored, by the same rule check_csrf_removal.py already used.

## 5. gate-50: a false positive and a false negative in the same regex

  FALSE NEGATIVE  the app-id argument had to be a QUOTED STRING, so every read
                  written the fleet-standard way — `getValueString(
                  Application::APP_ID, 'listing_register', '')` — was invisible.
                  Identical code with `'larpingapp'` FAILED. Same family as
                  #184. 7 security-relevant reads across 5 repos sit behind a
                  constant today.
  FALSE POSITIVE  the empty-compare guard required a closing paren immediately
                  after the empty string, so the correct compound guard
                  `if ($reg === '' || $sch === '')` was reported as unguarded —
                  twice, on code the gate was asking for. A guard that is a
                  boolean `return` rather than an `if` was rejected too.

Both directions are now asserted, including the opencatalogi#86 shape that
mixes them: one read guarded, the next unguarded two lines later.

## 6. gate-53 did not block the PR that creates larpingapp#286

Reintroducing #286 exactly — the check-in tab deleted from src/manifest.json,
`EventRoster` left registered in src/registry.js — reported PASS. Direction 1
of the registry cross-reference stays advisory for LEGACY orphans, correctly:
the gate cannot tell "wire it" from "delete it". But when the DIFF ITSELF
removed the last reference it can, and that finding now blocks. Pre-existing
orphans are untouched (larpingapp carries one today), so this is prevention,
not a burn-down list nobody can close.

## Verified working, repaired nothing

gate-46 (dangling file, dangling fragment, valid anchor), gate-48 (short and
fully-qualified attribute removal; a comment reword correctly stays green),
gate-49, gate-51 (title, description and nested items.properties independently),
gate-52's ratchet (growth fails, shrink passes), gate-54 (flat, nested and
$ref-carrying), gate-55.

## Deliberately NOT enforced

`title == key` on a schema property is a real gate-51 defect — the renderer
uses `prop.title || key`, so the user sees the raw technical key. Measured
across 10 repos: 148 occurrences, ALL of them in softwarecatalog, where they
are VNG-standardised element names (`identifier`, `type`, `name`) that must
not be renamed. Enforcing it would produce 148 findings with no legitimate end
state in the one repo that has them. Reported rather than gated (#252).

## Divergence to reconcile

gate-45 now answers an empty in-scope set with `na`; gate-40 answers it with
PASS, by a deliberate choice in #272 that cited the invariant test this PR
reworks. The invariant now discriminates on the REASON — the applicability
table's own phrasing must not appear once its prerequisite holds — so both
behaviours are expressible. The family should pick one.

## Testing

New: hydra-gates/scripts/lib/test_gate_45_to_55_acceptance.sh — 31 arms across
six families, discovered by run-helper-suites.sh. Against the package as
merged on main it fails 20 of 31; the 11 that pass are exactly the
anti-widening and no-regression controls. Every mutation asserts its anchor is
present before it plants.

Repos used, chosen for different shapes: larpingapp (register-owning,
manifest-driven, ships registry.js), nldesign (PHP templates, no .vue, no
register), doriath (ships no phpcs SpecTagSniff — the #246 control, held at
81 findings across 46 targets before and after the plant), openconnector
(41 register files).

Full package suite: 52 discovered suites pass, 2 quarantined as documented;
60/60 entry-point invariants.

### DIFF
--- a/hydra-gates/scripts/lib/check_custom_widget_ratchet.py
+++ b/hydra-gates/scripts/lib/check_custom_widget_ratchet.py
@@ -47,8 +47,31 @@ computed (there is no base to compare against).
 
 Output: one location-prefixed finding block per violation plus one
 ``[custom-widget-ratchet] base=N head=M delta=±K`` report line whenever the
-ratchet is computed. Exit code is the number of findings (capped at 99);
-the calling gate uses it as the failure count.
+ratchet is computed, and — ALWAYS, on every successful run — one
+
+    [custom-widget-ratchet] findings=N
+
+line. THAT line is the answer, not the exit status.
+
+WHY THE COUNT IS ON STDOUT AND NOT IN THE EXIT CODE (#209)
+----------------------------------------------------------
+This helper used to return its finding count as its exit status and the gate
+read it as the failure count. An exit status is ONE BYTE, and it is also how
+the interpreter reports that the helper never finished. Both meanings arrived
+on the same channel, so they were not distinguishable:
+
+  * a Python traceback exits 1, and the gate reported it as
+    ``FAIL — 1 custom-widget finding(s)``. Measured 2026-08-08 by injecting a
+    ``raise`` into ``main()``: the gate produced a plausible, actionable,
+    entirely fictional finding, and a reader chasing it would have gone
+    looking for a widget that does not exist.
+  * the count was clamped to 99 to stay inside the byte, so any run with 100+
+    findings under-reported — the same lossy-channel shape that made gate-19
+    report 266 findings as 10.
+
+Exit status is now a plain boolean (0 = clean, 1 = findings present). A gate
+that sees a non-zero exit WITHOUT a ``findings=`` line knows the helper died
+and must report WIRING, never a finding.
 """
 
 import os
@@ -355,12 +378,20 @@ def _changed_and_deleted(base_ref, candidate_paths):
 # --------------------------------------------------------------------------
 # Main.
 # --------------------------------------------------------------------------
+FINDINGS_LINE = "[custom-widget-ratchet] findings={n}"
+
+
 def _emit(findings, counts_line=None):
     if counts_line is not None:
         print(counts_line)
     for finding in findings:
         print(finding)
-    return min(len(findings), 99)
+    # The count, on stdout, unclamped — see the module docstring (#209).
+    # Printed LAST so it cannot be confused with a finding block's own text,
+    # and printed on EVERY successful run including the clean one, so its
+    # absence means "the helper did not finish".
+    print(FINDINGS_LINE.format(n=len(findings)))
+    return 1 if findings else 0
 
 
 def _format_delta(delta):
@@ -392,7 +423,9 @@ def main(argv):
         # ratchet needs a base to compare against and is not computed.
         head_count = sum(len(es) for es in head_customs.values())
         if head_count == 0:
-            return 0
+            # Still emit `findings=0` — the caller uses the PRESENCE of that
+            # line to tell a clean run from a helper that died (#209).
+            return _emit([])
         findings = [
             f"{path}:{e.line}: " + JUSTIFICATION_MSG.format(key=e.key)
             for path, es in sorted(head_customs.items())
@@ -433,7 +466,7 @@ def main(argv):
         for path in sorted(changed_head | base_only)
     )
     if not active:
-        return 0
+        return _emit([])
 
     # Per-app counts: unchanged files contribute identically to both sides.
     head_count = sum(len(es) for es in head_customs.values())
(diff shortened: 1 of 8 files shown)


rubenvdlinde:
## Correction to a number in this PR's description

The description says the constant-app-id blind spot was measured at **"7 security-relevant reads across 5 repos"**. That is an undercount, and the cause is mine: the loop I measured with enumerated ten repos by hand and **pipelinq and hermiq were not in the list**. A hand-written repo list decays exactly like a hand-written paths filter — the omission is invisible.

Recounted over twelve repos:

```
openconnector       1
procest             1
larpingapp          2
softwarecatalog     1
decidesk            2
pipelinq          291        <- omitted from the original count
```

pipelinq writes `Application::APP_ID` throughout, so **its entire gate-50 surface was invisible**: the gate reported PASS on that repo and now reports 38 findings. Spot-checked both ways:

* **real** — `lib/Service/ContactImportService.php:65` reads `client_schema` with no guard and passes it straight to `saveAndSerialize()`. An empty schema id there is an unscoped write; this is the opencatalogi#86 shape the gate exists for.
* **read-out** — `lib/Service/ActivityTimelineService.php:125-129` are entries in an array literal in `getConfig()`, the same shape #282 removed for variable app-ids. About 4-6 of the 38 look like this. They are not excluded, because they use a class constant and the gate cannot yet tell a scope decision from a read-out — a limitation now stated in the helper's own comment and pinned by test arms C9/C10.

So the direction of the fix is unchanged and the finding it recovers is real, but **"7 across 5 repos" should read "~298 reads across 6 repos, of which 38 are unguarded — 38 of them in pipelinq alone."** Recording it here rather than leaving the smaller number to be quoted onward.

Related: the sweep I used to validate this PR compared **verdicts**, not counts, and that is how a 23 -> 64 regression on softwarecatalog reached `main` — FAIL before, FAIL after. Fixed in #282; the method note is in that PR's description.

PR #286, opened 2026-08-09
fix(gate-54): a genuine cross-app reference could not be authored at all

rubenvdlinde:
## The gate was unclosable

`x-external-register` appears **zero times** in `check_relation_dialect.py`. The checker has no concept of a reference to a schema owned by another app — and for such a property **both arms fail**:

| Authoring | Verdict |
|---|---|
| `"$ref": "case"` | (f) `$ref 'case' does not resolve to a schema key in the register set (case-exact)` |
| no `$ref` | (b) `relation-shaped property (format:uuid + relation description) lacks canonical $ref (ADR-062 rule 7)` |

Reproduced directly against `main` (`e7bde0a`), one finding on each arm.

**Measured on docudesk**: `correspondence.caseReference` and `generatedDocument.zaakId` both reference a Zaak that lives in **Procest's** register, and both already declare `x-external-register: procest`. There is no authoring that satisfies the gate. The only route to green was to reword the description until `_RELATION_DESC_RE` stopped matching — **degrading documentation to dodge a regex**, which is exactly what a gate must never reward. This is gate-59's own subject matter.

## The rule

OpenRegister resolves `$ref` inside **one** register set, so a foreign schema is not expressible as a relation. A property carrying `x-external-register` is therefore a **plain identifier**: it must not be asked for a `$ref`, and it must not carry one.

## The exemption is narrow, and the true positive survives

- rule **(b)** skips a property that declares `x-external-register`;
- rule **(f)** still reports one that declares **both** `x-external-register` and a `$ref` — that `$ref` is one OpenRegister can never resolve. The message now names the actual fix (**"drop the `$ref`"**) instead of the generic "does not resolve", which pointed the author in the wrong direction and is what produced the flip-flop in the first place;
- a dangling `$ref` **without** the marker keeps its original message;
- filter-token validation is untouched.

## Tests

Four added to `test_check_relation_dialect.py`, each failing before this change:

- `test_external_reference_without_ref_is_accepted` — the closable arm
- `test_external_reference_with_a_dangling_ref_is_still_reported` — true positive survives, message is actionable
- `test_a_local_dangling_ref_is_unaffected` — control: widening a checker until it catches nothing is not a fix
- `test_external_marker_does_not_excuse_a_bad_filter_token` — control: the exemption is scoped to the `$ref` rules only

Full helper suite: **28 tests, all green** (24 pre-existing + 4 new). Verified the 2 pre-fix failures with `python3 -m unittest` before implementing.

## End-to-end

docudesk `gate-54: FAIL — 2 findings` → **`PASS`**, with its register also corrected to drop the two dangling `$ref`s (ConductionNL/docudesk, separate PR). Both halves were required: the register fix alone flipped the finding to the *other* arm.

### DIFF
--- a/hydra-gates/scripts/lib/check_relation_dialect.py
+++ b/hydra-gates/scripts/lib/check_relation_dialect.py
@@ -338,6 +338,33 @@ def _global_schema_keys(paths):
 _RELATION_DESC_RE = re.compile(r"\b(reference to|verwijzing naar|uuid of the|fk to)\b", re.I)
 
 
+def _is_external_ref(prop):
+    """True when a property points at a schema in ANOTHER app's register.
+
+    OpenRegister resolves ``$ref`` inside ONE register set, so a schema owned by
+    a different app is not expressible as a relation at all. Registers mark such
+    properties with ``x-external-register: <app>`` and carry the bare identifier
+    (``type: string`` + ``format: uuid``).
+
+    Without this notion the gate was UNCLOSABLE for a genuine cross-app
+    reference — measured 2026-08-09 on docudesk, whose `correspondence.
+    caseReference` and `generatedDocument.zaakId` point at a Zaak in Procest:
+
+      WITH    ``"$ref": "case"``  -> check (f) "does not resolve to a schema key
+                                     in the register set (case-exact)"
+      WITHOUT ``$ref``            -> check (b) "relation-shaped property ...
+                                     lacks canonical $ref (ADR-062 rule 7)"
+
+    Both arms failed, so the only route to green was rewording the description
+    until ``_RELATION_DESC_RE`` stopped matching — degrading documentation to
+    dodge a regex. The exemption is deliberately narrow: it suppresses ONLY the
+    two ``$ref`` rules, and a property that declares ``x-external-register``
+    *and* a ``$ref`` is still reported, because that ``$ref`` is one
+    OpenRegister can never resolve.
+    """
+    return isinstance(prop, dict) and bool(str(prop.get("x-external-register") or "").strip())
+
+
 def _ref_of(prop):
     """Return (raw_ref_value, is_array) for a relation property, or (None,_).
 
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
**Independent confirmation from a second repo and a different cross-app shape — this fixes openconnector too.**

I hit the same wall on openconnector and filed #289 before seeing this. Same working tree, two package refs, nothing changed but the checker:

| package ref | gate-54 |
|---|---|
| `c51a225` / `651e5c5` (main) | **FAIL** — `lacks canonical $ref (ADR-062 rule 7)` |
| `7f5f9e7` (this PR) | **PASS** |

The property is `source.configuration.authentication.credentialRef.credentialId` in `lib/Settings/openconnector_register.json`, annotated `x-external-register: openregister` with no `$ref`.

### Why this is a useful second data point

docudesk's cases point at a **Zaak in Procest**. Mine points at a **credential in OpenRegister's credential broker**:

- different owning app,
- different reason for the boundary — the secret must never be stored on the source, so openconnector deliberately holds only the identifier,
- and it is nested three levels deep (`configuration` → `authentication` → `credentialRef` → `credentialId`) rather than sitting at the schema root.

`_is_external_ref()` handles it unmodified. Keying the exemption on the **property** rather than on the register's shape is what makes that work, and it is the right axis.

### The narrowness holds

I confirmed the exemption suppresses only the two `$ref` rules — a property carrying **both** `x-external-register` and a `$ref` is still reported, which is correct, since that `$ref` is one OpenRegister can never resolve. The true positive is preserved.

### One note for whoever merges this

Until this lands, the annotation is inert on main and the gate reports the original finding byte-for-byte — so a repo that adopts `x-external-register` early sees no change and might reasonably conclude the key does nothing. `grep -c x-external-register check_relation_dialect.py` is **0** on `651e5c5`. I am landing the annotation on openconnector anyway, because the property genuinely *is* a foreign-register reference and documenting that is correct regardless; openconnector's gate-54 will go green the moment this merges, with no further change on my side.

#289 can be closed as a duplicate of this once it merges.

Issue #304, opened 2026-08-09
gate-55's hardcoded icon mirror has drifted from nc-vue: it accepts ClipboardList (renders '?') and rejects ClipboardListOutline (correct)


rubenvdlinde:
## Summary

`check_detail_page_discipline.py` hardcodes `ICON_REGISTRY` as a mirror of nc-vue's `src/components/CnWidgetGrid/widgetIcons.js`. The mirror has drifted by one entry, and it drifted in the direction that makes the gate **wrong in both directions at once**.

| | in nc-vue registry | in gate mirror |
|---|---|---|
| `ClipboardList` | ❌ no | ✅ yes |
| `ClipboardListOutline` | ✅ yes | ❌ no |

So gate-55 **accepts** `ClipboardList`, which renders the `?` fallback in the running app, and **rejects** `ClipboardListOutline`, which is the correct name.

## Measured

Against `@conduction/nextcloud-vue` as installed (pipelinq's `node_modules`):

```
real registry count: 55
mirror count: 55
real NOT in mirror: ClipboardListOutline
mirror NOT in real: ClipboardList
```

Extracted from the real file by its import statements (`vue-material-design-icons/<Name>.vue`), which is how the registry is actually built — each icon is imported individually, no barrel.

## Why it matters beyond one name

The gate's own comment says the mirror is *"a hardcoded mirror — when the registry changes, replace the names below"*. Nothing enforces that, so the mirror is a **silent expiry date**, the same shape as the pinned-callee problem. There is no test asserting mirror == registry.

## There is a second registry, and gate-55 does not know about it

An icon must be in **two** places to render:

1. **nc-vue's** `widgetIcons.js` — what gate-55 checks.
2. **The app's own** `src/icons.js` (ADR-077) — what **gate-60** checks.

An icon in (1) but not (2) renders with **no icon at all** — worse than the `?` fallback. Measured on pipelinq: `Timeline` is in nc-vue's registry *and* in gate-55's mirror, but not in pipelinq's `src/icons.js`. gate-55 was **green** on it; only gate-60 caught it.

⚠️ And gate-60 is easy to lose: it reports `SKIPPED (wiring)` when `vue-material-design-icons` is absent, which is the default in a fresh worktree without `npm ci`. In that state gate-55 is the only icon check running, and it is the one with the stale mirror. A broken icon ships under a green board.

## Suggested fix

1. Re-sync the mirror (`ClipboardList` → `ClipboardListOutline`).
[...]

PR #324, opened 2026-08-10
fix(gate-55): add the missing Map key to the icon mirror, and pin its derivation

rubenvdlinde:
Addresses #304 — but **not in the direction that issue proposes**. Please read the correction below before merging.

## What #304 got right

The mirror is a hardcoded copy with nothing enforcing it, so it is a silent expiry date. That part stands, and this PR closes it with a test.

## What #304 got wrong

#304 reports:

| | in nc-vue registry | in gate mirror |
|---|---|---|
| `ClipboardList` | ❌ no | ✅ yes |
| `ClipboardListOutline` | ✅ yes | ❌ no |

…and proposes swapping `ClipboardList` → `ClipboardListOutline`.

**That would invert the gate in both directions.** `CnIcon` resolves a name by **key lookup**:

```js
_registry[name] || DASHBOARD_ICONS[name] || ... || HelpCircleOutline
```

so what matters is the **object key** in `DASHBOARD_ICONS`. Two keys deliberately differ from the file they import — the key is the concept, the file is just the glyph chosen for it:

```js
ClipboardList: ClipboardListIcon   // <- 'vue-material-design-icons/ClipboardListOutline.vue'
Map:           MapIcon             // <- 'vue-material-design-icons/MapOutline.vue'
```

#304 extracted the registry "by its import statements", which is where the inversion came from. `ClipboardList` **is** a valid key and renders correctly; `ClipboardListOutline` is **not** a key and renders `?`. Applying the suggested swap would make the gate reject the only working name and accept the broken one.

Measured against the real registry keys on both nc-vue lines (`beta` and `feat/vue-3`, 56 keys each) and against pipelinq's installed `1.0.0-beta.197` — all three carry `ClipboardList:` **and** `Map:` as keys:

```
mirror count: 55 | real KEY count: 56
real NOT in mirror: ['Map']
mirror NOT in real: []
```

## The one genuine drift

`Map`, added upstream with the Map dashboard widget. Without it the gate **rejects a valid icon**. Latent today — no app currently uses it — which is exactly why it would have surfaced as a baffling gate failure on the first one that did.

## The test

`test_check_detail_page_icon_registry.py`, picked up automatically by `run-helper-suites.sh` (no workflow edit). It pins **both** halves:

1. the working **keys** are present;
2. the **filename-derived** spellings are absent — so a future "resync" done by grepping imports fails here loudly instead of silently inverting the gate, which is the specific mistake this PR is correcting;
3. a **positive control**, so a mirror that had accidentally become "everything" cannot satisfy the other two.

Proof it can fail — against `origin/main`:

```
AssertionError: 'Map' not found in {...} : 'Map' is a DASHBOARD_ICONS key;
without it the gate rejects an icon that renders correctly
FAILED (failures=1)
```

With this change: `Ran 3 tests — OK`.

## Also measured while here

The related report that **decidesk has 28 icons rendering the `?` fallback does not reproduce.** decidesk's manifests use **101 distinct icon values across 330 occurrences, and every one resolves** — its own `src/icons.js` registers exactly those 101 (ADR-077, landed in decidesk#363/#364), and `CnIcon` falls back only when a name is in *neither* `registerIcons()` *nor* `DASHBOARD_ICONS`. Both gates confirm it:

```
gate-55 (detail-page-discipline) on decidesk : exit 0, no findings
gate-60 (icon-vocabulary)        on decidesk : checked 49 manifest(s): 0 failure(s), 0 warning(s)
```

So no icons need adding to nc-vue's registry; this one-key mirror correction is the whole of the real defect.

#304's third point still stands and is worth doing separately: a green gate-55 should not be read as "this icon renders", because app-level registration is gate-60's job.

### DIFF
--- a/hydra-gates/scripts/lib/check_detail_page_discipline.py
+++ b/hydra-gates/scripts/lib/check_detail_page_discipline.py
@@ -56,6 +56,30 @@ import sys
 # This list is a hardcoded mirror — when the registry changes, replace the
 # names below (kept as a single, alphabetically-ordered constant on purpose so
 # the swap is a trivial one-line edit).
+#
+# ⚠️ MIRROR THE OBJECT KEYS, NOT THE IMPORT FILENAMES.
+#
+# `CnIcon` resolves a name with a plain key lookup:
+#
+#     _registry[name] || DASHBOARD_ICONS[name] || ... || HelpCircleOutline
+#
+# so what must appear here is the KEY on the left of each entry in
+# `DASHBOARD_ICONS`. Two keys deliberately differ from the file they import,
+# because the registry name is the concept and the file is just the glyph
+# chosen for it:
+#
+#     ClipboardList: ClipboardListIcon   <- 'vue-material-design-icons/ClipboardListOutline.vue'
+#     Map:           MapIcon             <- 'vue-material-design-icons/MapOutline.vue'
+#
+# Deriving this set by grepping the import statements therefore produces
+# `ClipboardListOutline` / `MapOutline` — names that are NOT registry keys and
+# that render the "?" fallback — while dropping the two that work. That mistake
+# is what .github#304 reported as a registry drift; the mirror was right about
+# `ClipboardList` and the measurement was wrong. Do not "fix" it in that
+# direction.
+#
+# `Map` was the one genuine omission (added by nextcloud-vue's Map dashboard
+# widget); without it the gate REJECTED a valid icon.
 # --------------------------------------------------------------------------
 ICON_REGISTRY = {
     "Account", "AccountBoxOutline", "AccountGroup", "AlertCircleOutline",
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
One more measurement while verifying the mirror, which strengthens #304's point (3) — **not fixed here**, flagging it rather than widening this PR.

gate-60's `semantic-icons.json` prescribes the **filename** spellings, exactly the ones gate-55 rejects:

```
Map                    in gate-60 vocab: False
MapOutline             in gate-60 vocab: True
ClipboardList          in gate-60 vocab: False
ClipboardListOutline   in gate-60 vocab: True
```

Both gates are internally correct, because they check **different resolution paths**:

- gate-60 governs nav/menu icons, which resolve through the app's own `registerIcons()`. `MapOutline.vue` and `ClipboardListOutline.vue` exist upstream, so a registered `MapOutline` renders fine.
- gate-55 governs widget icons against nc-vue's `DASHBOARD_ICONS`, whose keys are `Map` / `ClipboardList`.

The consequence is that the two gates give an author **opposite advice for the same glyph**, and worse: `CnIcon` tries the app registry *first* —

```js
_registry[name] || DASHBOARD_ICONS[name] || ... || HelpCircleOutline
```

— so an app that registers `MapOutline` renders a widget icon named `MapOutline` correctly, while **gate-55 still rejects it**. That is a false positive in the opposite direction from the one #304 was worried about, and it comes from the same root cause: gate-55 models only the library vocabulary and cannot see the app's registry.

So #304's point (3) is worth more than a wording change — gate-55's verdict is only sound in one direction. Left as-is here so this PR stays the one-key correction it claims to be.

Issue #194, opened 2026-08-06
quality.yml: gating the test tier on the security tier silently deletes all test evidence (CVE-2026-67434 disabled PHPUnit/Newman/E2E fleet-wide today)


rubenvdlinde:
## Summary

`quality.yml` gates the **test** tier on the **security** tier:

```yaml
if: ${{ ... && !cancelled() && needs.security.result != 'failure' }}
```

Applied to `phpunit` (L1464), `newman` (L1720), `playwright` (L1985), `journeydoc-capture` (L2739) and `sbom` (L3999).

The consequence: **any** advisory against **any** dependency — including a dev-only code formatter with no runtime exposure — turns PHPUnit, Newman and E2E into `skipped` across the entire fleet, simultaneously.

**A skipped job renders as a grey tick, not a red X.** The run does not look broken. It looks fine. The tests simply did not run.

This is not hypothetical — it happened this morning across 16 repos, and it fooled an agent into filing a false "fully green" report.

## Timeline (2026-08-06)

| time (UTC) | event |
|---|---|
| 2026-08-05 23:53 | `CVE-2026-67434` published against `squizlabs/php_codesniffer <3.13.6` (OS command injection, [GHSA-hmqg-cxww-wqhq](https://github.com/PHPCSStandards/PHP_CodeSniffer/security/advisories/GHSA-hmqg-cxww-wqhq)) |
| 2026-08-06 00:17 | upstream ships `3.13.6` |
| ~08:48 | fleet still green — advisory not yet served by the audit path. nldesign baseline runs **110 E2E tests** |
| **~11:15** | advisory begins being served. `Security (composer)` starts failing fleet-wide |
| ~11:15 onward | `pipelinq`, `procest`, `decidesk`, `openbuild`, `openregister` observed with `Security (composer)=failure` → test tier `skipped` |

The trigger was **not** a code change in any repo. `composer audit` queries the **live** Packagist advisory database, so the fleet's test coverage was deleted by an external event, with no commit and no signal.

Note this is independent of the `roave/security-advisories` lockfile pin — that is a separate mechanism (an install-time conflict rule); `composer audit` hits the network every run.

## Why this shape is dangerous

1. **It fails silent.** `skipped` is not `failure`. Dashboards, `gh pr checks` summaries, and any tally that counts failures read zero and report success. A skipped test job is **not a pass — it is the absence of a verdict**, but it is rendered in the same colour family as a pass.
[...]

PR #326, opened 2026-08-10
ci(quality): give forced runs a real verdict, install the app's node deps in hydra-gates, cache everywhere

rubenvdlinde:
## What this changes

Four changes to the shared `quality.yml` that all 16 fleet repos consume, plus the removal of four `needs:` edges that gated nothing.

### 1. `hydra-gates` installs the app's node dependencies (fixes a live fleet-wide red)

gate-60 (`icon-vocabulary`) resolves `vue-material-design-icons` from `<app>/node_modules` to answer one question: *does this manifest icon name exist upstream, or was it invented?* The job installed only the gates package's own `ajv`, so that directory never existed, gate-60 reported `SKIPPED (wiring)`, and `--require-full-coverage` turned that into a job failure in every repo.

**A/B on one tree, differing only in whether `node_modules` is present.** Subject: a planted true positive — a menu icon named `LedgerOutline`, an MDI component with no upstream existence, which renders blank anywhere it is copied.

| | gate-60 | exit |
|---|---|---|
| deps **absent** (main today) | `SKIPPED (wiring)` → `ALL GATES PASSED — EXCEPT GATES 60, WHICH DID NOT RUN` | 98 |
| deps **present** (this PR) | `FAIL — 1 icon(s) outside the canonical vocabulary (ADR-077)` | 1 |

Read the failing arm carefully: today the job **does** go red, but it goes red about *coverage* and never names the invented icon. The finding was not merely unreported — it was unobtainable. Removing the plant on the same tree returns exit 0 and `gate-60: PASS`, so the gate is shown able to answer in **both** directions, not merely to stop complaining. Coverage moves 27/28 → 28/28 applicable.

Deliberately **not** a `download-artifact` from `frontend-build`: an artifact-dependent job *skips* when its producer fails, and a skipped job renders as a grey tick.

Guarded — a repo with no `package.json`/`package-lock.json` announces the reason and continues. Verified live: `.github` has no `package.json`, and the run printed the notice and proceeded.

### 2. ~~`hydra-gates` runs in parallel~~ — WITHDRAWN, and the reason matters

This PR originally removed `needs: [playwright]`. **It has been restored.** Recording the reversal rather than quietly dropping it, because both halves of the original reasoning failed in instructive ways.

The edge exists only to hand `tests/axe/report.json` to gate-33. The case for removing it was a survey of all 20 consumers on both `main` and `development`: zero set `enable-axe`, eleven with explicit comments saying so deliberately, and this file's own input docs record gate-33 as `SKIPPED` in every repo since it was written. So it looked like pure serialisation — a 31s job behind `php-quality → frontend-build → playwright`.

**Two things killed it.**

1. **The survey was a snapshot, not a property.** [docudesk#423](https://github.com/ConductionNL/docudesk/pull/423) is open to set `enable-axe: true`. "No consumer does X" is a fact with an expiry date; a structural decision built on one expires with it, silently, in someone else's repository.

2. **The replacement was worse than the problem — measured, not argued.** With the edge gone, an `enable-axe` caller cannot receive the artifact, so the job refused the configuration up front. That refusal sat before the checkout:

| | `[gate-*]` lines emitted |
|---|---|
| refusal (edge removed) — [run 31370581318](https://github.com/ConductionNL/.github/actions/runs/31370581318) | **0** |
| edge restored — [run 31370809231](https://github.com/ConductionNL/.github/actions/runs/31370809231) | **62** |

gate-33 produced no verdict **and neither did the other 63 gates**. It did not `_skip` with a named reason and it did not `_pass` — the whole job died. That is exactly what the note on the `axe-check` step already forbids: *a gate that suppresses sixty other gates to report itself is its own kind of outage*, which is why the axe failure is recorded there and raised at the **end**.

**What survives.** `php-quality` is still out of the `needs:` of `phpunit`, `newman` and `playwright`. Playwright starts earlier, so the gates verdict lands around **t+4.8m instead of t+7m** — a real improvement with no new machinery and no path that has never been exercised.

**How to get full parallelism later.** `needs:` cannot be conditional in GitHub Actions, so it needs two mutually exclusive jobs — which today means duplicating ~120 lines of a security-critical runner, and two copies that can drift is the failure mode this file exists to document. Extract the steps into a composite action referenced as `ConductionNL/.github/.github/actions/hydra-gates@main` first (a bare `./` path in a reusable workflow resolves against the **caller's** checkout, not this repo). Separate change, own testing.

### 3. A forced run produces a verdict

`workflow_dispatch` has no PR target and no `event.before`, so the base resolved to HEAD and the run exited 99 `NOTHING WAS CHECKED`. A dispatch is now **full-repo scope**, and the scope *and its reason* are printed and written to the step summary.

**Proved on three real dispatch runs of this branch, in both directions:**

| run | tree | result |
|---|---|---|
| [31366747441](https://github.com/ConductionNL/.github/actions/runs/31366747441) | as-is | `SCOPE FOR THIS RUN: full` · `Base ref: n/a — --full requested` · `COVERAGE: 4 of 64` · `RESULT: 1 GATE(S) FAILED` · **zero occurrences of "NOTHING WAS CHECKED"** |
| [31366930475](https://github.com/ConductionNL/.github/actions/runs/31366930475) | + planted conflict markers | `[gate-21] conflict-markers: FAIL — 1 file(s)` · `RESULT: 2 GATE(S) FAILED` |
| [31367009041](https://github.com/ConductionNL/.github/actions/runs/31367009041) | plant removed | `[gate-21] conflict-markers: PASS` · back to `1 GATE(S) FAILED` |

The constant failure across all three is `gate-4 composer-audit` — a genuine pre-existing finding about this repo (no `composer.lock`), not an artefact of the change.

`--full` is safe at **any** pin: verified against every published tag, `v1.0.0` through `v1.6.0`, all of which parse it. It therefore needs none of the capability probing `--axe-enabled` and `--require-full-coverage` require.

**It cannot pass over nothing.** On a forced run the `COVERAGE` count is read; zero gates reporting a result — *or a missing COVERAGE line* — fails the job with a named reason. Shown able to fail in both of those directions and to pass only on a real count.

**Behavioural delta is minimal and precise:** `pull_request` is unchanged, and `push` is unchanged (both still diff-scoped, same base). Only `workflow_dispatch`/`schedule` — which previously could not produce a verdict at all — behave differently.

### 4. Caching, and vestigial `needs:` removed

npm download caches on the Newman, Playwright, journeydoc and SBOM legs; composer download caches on every leg that runs composer. `actions/cache` is used rather than `setup-node`'s `cache: npm` wherever a lockfile may be absent, because that option **hard-fails** without one — a trap this file already carries two scars from.

**Latent staleness bug fixed:** `hashFiles('composer.lock')` returns `''` when the file is absent, so on the three fleet repos with no lockfile (hermiq, hrmq, zaakafhandelapp) the key collapsed to the constant `Linux-composer-`. An Actions cache entry is **immutable once written**, so those repos restored the same `vendor/` forever and no dependency change could rotate it — a cache that cannot expire is a stale tree with a fast install. `composer.json` now joins the key. Verified live: the key resolved to `Linux-composer-files-be9703db…`, a real content hash derived from `composer.json` alone.

`php-quality` removed from the `needs:` of `phpunit`, `newman` and `playwright`. **`security` is kept everywhere** — it resolves in seconds (9s and 16s for its two legs in the runs measured here), so removing it would buy no measurable time. `journeydoc-capture` is unchanged.

This file **already documented `php-quality` as "retained for ordering only"** — #194 and the 2026-08-04 procest incident established that gating the test tier on it let a phpmd *complexity* finding delete the entire tier, rendering as a grey "skipping" placeholder. The `if:` clause was removed then; the `needs:` entry was left behind. Ordering was the whole cost: `!cancelled()` suppresses the implicit `success()`, but it does not release a job to **start**.

Neither edge was ever a gate and neither is now. Verified no step anywhere reads `needs.php-quality.*` or `needs.security.*`. `playwright` keeps `frontend-build`, which it genuinely consumes via `needs.frontend-build.outputs.built`. **Security still blocks the merge**, unchanged, via `Quality Report`.

The complete `needs:` delta is four lines:

```
 phpunit      [php-quality, security]                 -> [security]
 newman       [php-quality, security]                 -> [security]
 playwright   [php-quality, security, frontend-build] -> [security, frontend-build]
```

`hydra-gates` keeps `needs: [playwright]` (see §2) and `journeydoc-capture` keeps `needs: [security]` — both unchanged.

## Honest note on elapsed time

This is about **removing false failures and unblocking parallelism**, not a big speed-up. Two things worth stating rather than glossing:

- **Caching pays off across runs, not within one.** Concurrent jobs all miss a cold cache; only jobs starting *after* a producer finishes get a hit. Serialising behind a build to "warm the cache" would be a net loss on wall-clock. Landed for its own sake and for the gate-60 fix.
- **I could not demonstrate a cache HIT, and here is why.** `.github` is not a PHP app, so every composer-bearing job fails on it, and `actions/cache` declares `post-if: success()` — a failing job never saves, so no second run can hit. What *is* demonstrated: the cache steps execute, and produce a real content-derived key rather than the constant-key collapse. The first fleet repo to run this on a green PR will show the hit.

## Verification

- `hydra-gates/tests/test-hydra-gates-bin.sh` — **61 passed, 0 failed**
- `hydra-gates/tests/run-helper-suites.sh` — **44 suites passed, 0 failed**
- YAML parses; every `needs:` target exists; no dangling `needs.*` context (the `needs.playwright.result` reference inside the gates job was removed with the edge)
- Backward compatibility against real consumers: **no `package.json`** (`hydra`, and `.github` itself — proven live, the guard announced and continued) and **no `composer.lock`** (hermiq, hrmq, zaakafhandelapp — the composer.json fallback proven live)
- Nested lockfiles (17 of 20 repos carry `docs/package-lock.json`; openregister also `tests/integration/`) are untouched — `frontend-setup-command` is unset in every consumer, so this workflow never installs them, and no cache keys on them

## Not done

- No bounded wait for the axe artifact (see §2 — untestable with zero consumers)
- `sbom → security` left in place: it is off the critical path and not part of this brief

### DIFF
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -63,7 +63,8 @@ jobs:
       enable-phpunit: false
       enable-newman: false
       enable-playwright: false
-      enable-hydra-gates: false
+      # TEMPORARY proof harness — REVERTED in the next commit.
+      enable-hydra-gates: true
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -63,7 +63,11 @@ jobs:
       enable-phpunit: false
       enable-newman: false
       enable-playwright: false
-      enable-hydra-gates: false
+      # TEMPORARY — proof harness for the forced-run scope change. REVERT TO
+      # false BEFORE MERGE. Turned on so a workflow_dispatch of this file
+      # actually materialises the Hydra Gates job and can be observed producing
+      # a verdict instead of "NOTHING WAS CHECKED".
+      enable-hydra-gates: true
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -57,7 +57,8 @@ jobs:
     uses: ./.github/workflows/quality.yml
     with:
       app-name: github-selftest
-      enable-php: false
+      # TEMPORARY — cache proof harness. REVERT TO false BEFORE MERGE.
+      enable-php: true
       enable-npm: false
       enable-frontend: false
       enable-phpunit: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -63,7 +63,9 @@ jobs:
       enable-phpunit: false
       enable-newman: false
       enable-playwright: false
-      enable-hydra-gates: false
+      # TEMP proof harness — reverted below.
+      enable-hydra-gates: true
+      enable-axe: true
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -57,18 +57,13 @@ jobs:
     uses: ./.github/workflows/quality.yml
     with:
       app-name: github-selftest
-      # TEMPORARY — cache proof harness. REVERT TO false BEFORE MERGE.
-      enable-php: true
+      enable-php: false
       enable-npm: false
       enable-frontend: false
       enable-phpunit: false
       enable-newman: false
       enable-playwright: false
-      # TEMPORARY — proof harness for the forced-run scope change. REVERT TO
-      # false BEFORE MERGE. Turned on so a workflow_dispatch of this file
-      # actually materialises the Hydra Gates job and can be observed producing
-      # a verdict instead of "NOTHING WAS CHECKED".
-      enable-hydra-gates: true
+      enable-hydra-gates: false
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -63,8 +63,7 @@ jobs:
       enable-phpunit: false
       enable-newman: false
       enable-playwright: false
-      # TEMPORARY proof harness — REVERTED in the next commit.
-      enable-hydra-gates: true
+      enable-hydra-gates: false
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -65,7 +65,6 @@ jobs:
       enable-playwright: false
       # TEMP proof harness — reverted below.
       enable-hydra-gates: true
-      enable-axe: true
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality-selftest.yml
+++ b/.github/workflows/quality-selftest.yml
@@ -63,8 +63,7 @@ jobs:
       enable-phpunit: false
       enable-newman: false
       enable-playwright: false
-      # TEMP proof harness — reverted below.
-      enable-hydra-gates: true
+      enable-hydra-gates: false
       enable-license-check: false
       enable-journeydoc-capture: false
       enable-coverage-guard: false
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -1603,29 +1603,38 @@ jobs:
     # and beta in all 25 fleet repos. Decoupling costs a few CI minutes on a PR
     # that cannot merge anyway; it buys a verdict that exists.
     #
-    # `needs:` USED to be retained "for ordering only" — and ordering was the
-    # entire cost. With `!cancelled()` and no result condition, this job already
-    # ran regardless of static-analysis colour and regardless of the advisory
-    # feed; the edges did not gate anything, they only made it WAIT.
+    # ── `php-quality` REMOVED from `needs:`; `security` KEPT ────────────────
     #
-    # They are now removed. `needs:` in GitHub Actions is a scheduling edge as
-    # well as a conditional one: `!cancelled()` suppresses the implicit
-    # `success()`, but it does not release the job to start. So PHPUnit sat
-    # behind php-quality for no verdict it consumed. Measured on openregister
-    # run 31358011759: php-quality 2m, and PHPUnit is the 4m job behind it.
+    # `needs:` in GitHub Actions is a SCHEDULING edge as well as a conditional
+    # one. `!cancelled()` suppresses the implicit `success()`, so this job
+    # already ran regardless of static-analysis colour — but suppressing the
+    # condition does not release the job to START. So PHPUnit sat behind
+    # php-quality for a verdict it never consumed.
+    #
+    # That php-quality is not load-bearing here is not an inference — it is
+    # recorded directly above by #194 and by the 2026-08-04 procest incident:
+    # gating on it meant a phpmd COMPLEXITY finding deleted the entire test
+    # tier, and the loss rendered as a grey "skipping" placeholder. The `if:`
+    # clause was removed then; the `needs:` entry was simply left behind, and
+    # the comment that survived called it "retained for ordering only".
+    # Ordering was the whole cost. Measured on openregister run 31358011759:
+    # php-quality 2m, with the 4m test jobs queued behind it.
+    #
+    # `security` STAYS. It resolves in seconds (9s and 16s for its two legs in
+    # the runs measured on this change), so it buys back no meaningful time,
+    # and keeping it preserves the conventional ordering with no latency cost.
+    # It is still NOT a gate — the `if:` reads no result from it, so the
+    # 2026-08-06 advisory shape (one CVE against a dev-only formatter deleting
+    # the fleet's test tier) cannot return through this edge.
     #
     # Nothing in this job's steps reads `needs.php-quality.*` or
-    # `needs.security.*` — verified across the whole file, where every surviving
-    # `needs.` expression belongs to the `report` job or to the Playwright job's
-    # frontend-build artifact check.
-    #
-    # Security still blocks the MERGE, exactly as before and by the same
-    # mechanism: `security` is in the Quality Report's `needs:`, and
-    # `quality / Quality Report` is the required context on main and beta in
-    # every fleet repo. Nothing is weakened; a job simply stops waiting.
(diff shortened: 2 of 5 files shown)


rubenvdlinde:
Self-review correction — a bug I introduced in this PR and then found, recorded here because the PR body above was written before it.

**Unescaped backticks in the node-deps guard warning.** Bash treats a bare backtick pair inside a double-quoted string as command substitution, so the message wrote ``npm ci`` as something to **execute** rather than something to name. The branch fires exactly when a repo has a `package.json` and no `package-lock.json` — i.e. precisely when the install is meant to be skipped — so it would have run `npm ci` in the one case that must not.

Demonstrated with the same shape before fixing:

```
echo "... and `echo INJECTED-COMMAND-RAN` requires a lockfile."
-> ::warning::... and INJECTED-COMMAND-RAN requires a lockfile.
```

**Why the three dispatch runs did not catch it:** `.github` has no `package.json` at all, so all three took the first branch. A green run over an unentered branch is not evidence about that branch. All three branches are now exercised directly:

| tree | output | `install` |
|---|---|---|
| no `package.json` | `::notice::No package.json at the app root…` | false |
| `package.json`, no lockfile | `::warning::… and ` + "`npm ci`" + ` requires a lockfile.` (literal) | false |
| both present | `Found package.json + package-lock.json at the app root.` | true |

Fixed in 7a1b7c6. Every other backtick in the file was already escaped; this was the only one that was not — grep for unescaped backticks in `echo` lines now returns only comments.

rubenvdlinde:
Correction pushed (fadd81e) — **`security` restored to the test tier's `needs:`; only `php-quality` is dropped.**

The earlier commit removed both. `security` resolves in seconds — **9s and 16s** for its two legs across the runs measured on this change — so removing it bought no measurable time while changing more of the graph than the goal required.

`php-quality` stays removed, and it is the edge the latency actually sat on. **This is not an inference.** #194 and the 2026-08-04 procest incident already established that gating the test tier on `php-quality` let a phpmd *complexity* finding delete the entire tier, rendering as a grey "skipping" placeholder. The `if:` clause was removed then; the `needs:` entry was left behind, and the surviving comment calls it *"retained for ordering only"*. Ordering was the whole cost — `!cancelled()` suppresses the implicit `success()`, but it does not release a job to **start**.

Neither edge was ever a gate and neither is now: no step in any of these jobs reads `needs.php-quality.*` or `needs.security.*`, and `security` still blocks the merge through `Quality Report`, unchanged.

The complete `needs:` delta in this PR is four lines:

```
-  phpunit             needs: [php-quality, security]              +  needs: [security]
-  newman              needs: [php-quality, security]              +  needs: [security]
-  playwright          needs: [php-quality, security, frontend-build]  +  needs: [security, frontend-build]
-  hydra-gates         needs: [playwright]                         +  (none)
```

`journeydoc-capture` keeps `needs: [security]` unchanged.

rubenvdlinde:
One behavioural consequence of full-repo scope worth stating before merge, because it is the thing a consumer will hit first.

**A forced run can exit 98 on gates that a PR run would classify `na`.** Full-repo scope is stricter than ADR-020 diff scope by construction: gates whose subject matter is genuinely absent report `NOT APPLICABLE` (fine, excluded from coverage), but two gates — **16 `spec-coverage` and 20 `or-objectservice-api`** — emit **no verdict line at all** on a `--full` run over a repo with no `lib/`. They appear only in the wrapper's set-difference as `GATES THAT DID NOT RUN: 16 20`, and with `--require-full-coverage` that is exit 98.

**This PR did not cause it and does not change the gate package** — `hydra-gates/scripts/run-hydra-gates.sh` is byte-identical between `main` and this branch. Reproduced on two independent trees (a synthetic fixture holding only a `README.md`, and `.github` itself live). Filed separately as #327.

It is also not a *silent* failure, which is the property that matters here: the run names the two gates and refuses to call itself green. That is the designed outcome — a forced run reports what it did **not** measure rather than passing over it.

**Checked, per the brief: this PR creates no new skip.** It *removes* one — gate-60's `SKIPPED (wiring)` becomes a real verdict once the app's deps are present, moving coverage 27/28 → 28/28 applicable on the fixture. `enable-axe` is false in all 20 consumers, so gate-33 keeps reporting `na` exactly as today.

rubenvdlinde:
**Correction — supersedes my two earlier comments in this thread.** `needs: [playwright]` on `hydra-gates` is **restored**. The `needs:` delta in this PR is now three lines, not four:

```
 phpunit      [php-quality, security]                 -> [security]
 newman       [php-quality, security]                 -> [security]
 playwright   [php-quality, security, frontend-build] -> [security, frontend-build]
```

`hydra-gates` and `journeydoc-capture` are unchanged.

**Why**, in two parts, both of which I got wrong the first time:

1. **My survey was a snapshot, not a property.** "Zero of 20 consumers set `enable-axe`" was true when measured and is being falsified right now by [docudesk#423](https://github.com/ConductionNL/docudesk/pull/423). Building a structural decision on a fleet-wide absence means the decision expires the moment someone fills it — silently, in a different repository.

2. **My replacement was worse than the problem, and only measuring showed it.** I described the `enable-axe` path as "a named refusal in seconds rather than a race". That description was wrong. The refusal sat before the checkout, so:

| | `[gate-*]` lines emitted |
|---|---|
| refusal, edge removed — [run 31370581318](https://github.com/ConductionNL/.github/actions/runs/31370581318) | **0** |
| edge restored — [run 31370809231](https://github.com/ConductionNL/.github/actions/runs/31370809231) | **62** |

It did not `_skip` with a named reason and it did not `_pass` — **the whole job died, taking all 63 other gates with it.** That is precisely what the note on the `axe-check` step already forbids, which is why that step records the axe failure and raises it at the *end*.

This is the same lesson as the backtick bug earlier in this PR, arriving through a different door: **`.github` sets `enable-axe: false`, so every green run I had was over a branch that was never entered.** A green run over an unentered branch is not evidence about that branch. I only found it by deliberately setting the input and looking at what came out.

**(D) re-proved on the final code**, with the edge restored, both directions: [31371012918](https://github.com/ConductionNL/.github/actions/runs/31371012918) `SCOPE: full` + `[gate-21] FAIL` + `2 GATE(S) FAILED`, then [31371128659](https://github.com/ConductionNL/.github/actions/runs/31371128659) `[gate-21] PASS` + back to 1.

The remaining latency win is real but smaller and honestly stated: dropping `php-quality` moves the gates verdict from ~t+7m to ~t+4.8m. Full parallelism needs a composite-action extraction first — noted in the job header as a separate change.

PR #332, opened 2026-08-10
fix(gate-19): score the Playwright config CI executes, not the one at the root

### LINKED ISSUE #331: gate-19: testIgnore awareness reads only the ROOT playwright.config.ts — a repo whose CI runs a SECOND config keeps the blind spot
Follow-up to #308 (gate-19 can now see `testIgnore`). The fix is real, but it does not close the hole for repos that run a config other than the root one — which is the shape the shared workflow explicitly supports.

## The gap

`check_e2e_coverage.py` resolves the config from the app root only:

```python
_PW_CONFIGS = ("playwright.config.ts", "playwright.config.js",
               "playwright.config.mts", "playwright.config.cjs")
...
for name in _PW_CONFIGS:
    cfg = app_dir / name
    if cfg.is_file():
        ...
        return
```

But `quality.yml` selects the config by `playwright-test-path` and falls back to the root:

```bash
CONFIG="${{ inputs.playwright-test-path }}/playwright.config.ts"
if [ ! -f "$CONFIG" ] && [ -f "playwright.config.ts" ]; then
  CONFIG="playwright.config.ts"
fi
npx playwright test --config="$CONFIG"
```

So when an app ships `<playwright-test-path>/playwright.config.ts`, **CI executes by that config while gate-19 judges by the root one.** The two can disagree completely.

## Measured on openregister

`playwright-test-path: tests/e2e/ci`, so CI runs `tests/e2e/ci/playwright.config.ts` (`testDir: '.'` — 4 files, 13 tests). The root config's `testDir` is the whole of `tests/e2e` (63 spec files), and gate-19 reads that one.

Result before my PR #2410: gate-19 reported **64 covered scenarios, and every single anchor backing them lived in a file CI never executed** —

```
21 tests/e2e/workflows/dsar-cases.spec.ts
20 tests/e2e/spec-coverage/register-i18n.spec.ts
18 tests/e2e/manifest-shell.spec.ts
12 tests/e2e/spec-coverage/saved-search-views.spec.ts
... (0 anchors in tests/e2e/ci/)
```

Zero anchors in the executed path. The gate's coverage number was entirely unexecuted, and #308's fix does not detect it because none of those files are `testIgnore`d by the root config — they are simply not the config CI runs.

## Suggested fix

Resolve the config the way the workflow does. The gate would need `playwright-test-path`, which it does not currently receive; simplest is to prefer `<path>/playwright.config.ts` when the path is passed (or when exactly one such config exists below `tests/`), falling back to the root as today.

A cheap partial improvement even without that input: if a config exists at a non-root location AND the root config's `testDir` covers files that config excludes, emit a warning naming both — silence there currently reads as agreement.

## Related, and not fixable by config parsing

Worth noting for anyone reading gate-19 numbers: a ref can also be dead without any skip at all. `dsar-cases.spec.ts` wraps assertions in

```js
if (await firstOpen.isVisible().catch(() => false)) { …assertions… }
```

with no `else`. On an unseeded instance the body runs nothing and the test PASSES, so `_ref_is_live` counts its refs. 15 of that file's 21 refs are in that shape — and separately, all 21 point at `openspec/specs/dsar-case-list/` and `openspec/specs/dsar-case-detail-actions/`, **neither of which exists** (they live only under `openspec/changes/dsar-case-ui/specs/` and were never synced).

rubenvdlinde:
Gate-19 parsed the **root** `playwright.config.ts`. The shared workflow runs a different file:

```bash
CONFIG="${{ inputs.playwright-test-path }}/playwright.config.ts"
if [ ! -f "$CONFIG" ] && [ -f "playwright.config.ts" ]; then
  CONFIG="playwright.config.ts"
fi
npx playwright test --config="$CONFIG"
```

A config under `playwright-test-path` **wins**. In every repo carrying both, the gate was scoring a suite CI never executes.

## Reproduction — openregister, both configs still present on `development`

| | `testDir` | spec files |
|---|---|---|
| root `playwright.config.ts` | `'./tests/e2e'` | **63** |
| `tests/e2e/ci/playwright.config.ts` | `'.'` → `tests/e2e/ci` | **4** |

`@e2e` anchors: **205 under `tests/e2e`, ZERO under `tests/e2e/ci`.** Every "covered" verdict gate-19 ever produced on that repo came from a file CI does not run.

Confirmed at the runner's own interface — invoked exactly as `run-hydra-gates.sh` invokes it, reading the **log**, not the exit code:

```
rc=1   stderr 0 bytes   859 findings
64 findings now cite tests/e2e/ci/playwright.config.ts by name
```

64 is the same number the fleet sweep found. No `Traceback`/`SyntaxError`/`AttributeError` anywhere in the 95 KB log — the interpreter lived.

## Two resolution bugs, both fixed

**1 — which file.** `playwright-test-path` is honoured: `$PLAYWRIGHT_TEST_PATH`, else the app's own caller workflow, else the shared workflow's default `tests/e2e`; then the same *exists-else-fall-back-to-root* rule. Read out of the caller workflow rather than plumbed through a new env var, so the gate agrees with the file that decides the behaviour — a second source of truth would drift exactly the way the root config drifted from the CI config. **`quality.yml` is untouched.**

**2 — what `testDir` means.** It is relative to the **config's** directory, not the repo root. openregister's CI config says `testDir: '.'` and means `tests/e2e/ci`. Reading it as repo-relative is what makes a 4-file suite look like a 63-file one. An absent `testDir` now means the config's own directory — Playwright's documented default.

Consequently *outside `testDir`* now means **not run**. That branch used to return "runs", on the reasoning that a mis-read `testDir` should never accuse. Sound while the gate read the root config, whose `testDir` spans everything; exactly wrong once the CI config is resolved.

## Fleet-wide finding count, measured before merging

Uncovered scenarios **6663 → 6785, +122**, in exactly **two** repos:

| repo | `playwright-test-path` | uncovered before | after | Δ |
|---|---|---|---|---|
| openregister | `tests/e2e/ci` | 796 | 860 | **+64** |
| hermiq | `tests/e2e/spec-coverage` | 433 | 491 | **+58** |

**Every other repo is unchanged.** This is the correct direction — a scenario "covered" by a test CI never ran is not covered. **hermiq was not in the original sweep** and is a second live instance of the same class; its canonical `development` carries both configs today.

Config resolution validated against all 20 real fleet repos: each resolves to the file the workflow would pick, and the six repos with a second config under the default `tests/e2e` (opencatalogi, openconnector, doriath, softwarecatalog, larpingapp, + doriath's) keep a `testDir` spanning `tests/e2e`, so their scenario counts do not move.

**No double-counting.** Repos with a single root config do not move at all. The collapse work in launchpad#85 and openregister#2410 is complementary — and a dedicated test arm pins that a *collapsed* repo still counts its anchors, so this stays correct whichever way those land.

## Both directions

11 new tests, 116 → **127**. Against the pre-fix helper:

| arm | pre-fix |
|---|---|
| anchor **outside** the executed config does NOT cover | **FAIL** (counted as covered) |
| anchor **inside** the executed config DOES cover | ok |
| repo **collapsed** to one root config still covers | ok |
| no caller workflow → falls back to root | ok |

The three must-not-change arms are green in **both** runs — the suite is not "everything fails". The remaining new arms exercise API that does not exist pre-fix (`_declared_test_path`, `config_rel`).

Also covered: quoted declarations (`"tests/e2e/spec-coverage"` — hermiq's spelling), commented-out declarations being ignored, `$PLAYWRIGHT_TEST_PATH` winning, and a config at the **default** path beating the root one.

`test_check_e2e_coverage.py` 127/127; full `run-helper-suites.sh` green.

Closes #331

### DIFF
--- a/hydra-gates/scripts/lib/check_e2e_coverage.py
+++ b/hydra-gates/scripts/lib/check_e2e_coverage.py
@@ -1199,21 +1199,121 @@ def _regexes(raw: str | None) -> list[re.Pattern[str]] | None:
     return out or None
 
 
+_WF_TEST_PATH_RE = re.compile(
+    r"^[^\S\n]*playwright-test-path[^\S\n]*:[^\S\n]*(['\"]?)([^'\"#\n]+)\1",
+    re.MULTILINE)
+
+
+def _declared_test_path(app_dir: Path) -> str:
+    """The `playwright-test-path` the app's CI actually declares.
+
+    Resolution mirrors how the value reaches the shared workflow:
+
+      1. ``$PLAYWRIGHT_TEST_PATH`` — set by a caller or a local run
+      2. the ``playwright-test-path:`` input in the app's own caller workflow
+      3. ``tests/e2e`` — the shared workflow's declared default
+
+    Read out of the caller workflow rather than plumbed through a new env var
+    on purpose: the gate has to agree with the file that decides the
+    behaviour, and a second source of truth would drift from it exactly the
+    way the root config drifted from the CI config.
+    """
+    env = os.environ.get("PLAYWRIGHT_TEST_PATH", "").strip()
+    if env:
+        return env.strip("/")
+    wf_dir = app_dir / ".github" / "workflows"
+    if wf_dir.is_dir():
+        for wf in sorted(wf_dir.glob("*.y*ml")):
+            try:
+                text = wf.read_text(encoding="utf-8")
+            except OSError:
+                continue
+            for m in _WF_TEST_PATH_RE.finditer(text):
+                value = m.group(2).strip().strip("/")
+                # `#` starts a YAML comment, so a commented-out or
+                # documentation line yields nothing usable. Several fleet
+                # workflows explain this input at length directly above it.
+                if value and not value.startswith("#"):
+                    return value
+    return "tests/e2e"
+
+
+def _resolve_config(app_dir: Path) -> Path | None:
+    """The config CI executes — NOT necessarily the one at the repo root.
+
+    THE GATE READ A DIFFERENT FILE THAN CI RAN (.github#331)
+    -------------------------------------------------------
+    The shared workflow does exactly this:
+
+        CONFIG="${{ inputs.playwright-test-path }}/playwright.config.ts"
+        if [ ! -f "$CONFIG" ] && [ -f "playwright.config.ts" ]; then
+          CONFIG="playwright.config.ts"
+        fi
+        npx playwright test --config="$CONFIG"
+
+    so a config under `playwright-test-path` WINS over the root one. This gate
+    read the root config unconditionally, so in every repo carrying both it
+    scored a suite CI never executes.
+
+    Measured 2026-08-10 on openregister, which declares
+    `playwright-test-path: tests/e2e/ci`:
+
+        root config   testDir './tests/e2e'  -> 63 spec files
+        CI config     testDir '.'            ->  4 spec files
+        @e2e anchors  205 under tests/e2e, and ZERO of them under tests/e2e/ci
+
+    Every "covered" verdict gate-19 ever produced on that repo came from a
+    file CI does not run.
+
+    `.ts` is tried first at both locations because that is the literal the
+    workflow builds; the other extensions are a fallback for repos that do not
+    go through the shared workflow at all.
+    """
+    sub = _declared_test_path(app_dir)
+    if sub:
+        cand = app_dir / sub / "playwright.config.ts"
+        if cand.is_file():
+            return cand
+    cand = app_dir / "playwright.config.ts"
+    if cand.is_file():
+        return cand
+    for name in _PW_CONFIGS:
+        cand = app_dir / name
+        if cand.is_file():
+            return cand
+    return None
+
+
 class _PlaywrightScope:
     """Answers: would ANY project run this test file?"""
 
     def __init__(self, app_dir: Path) -> None:
         self.parsed = False
         self.test_dir = ""
+        self.config_dir = ""
+        self.config_rel = ""
         self.projects: list[tuple[list[str] | None, list[re.Pattern[str]] | None]] = []
-        for name in _PW_CONFIGS:
-            cfg = app_dir / name
-            if cfg.is_file():
-                try:
-                    self._parse(_strip_ts_comments(cfg.read_text(encoding="utf-8")))
-                except OSError:
-                    return
-                return
+        cfg = _resolve_config(app_dir)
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
## Canonical fleet measurement (supersedes the local numbers in the body)

The body's per-repo figures came from local checkouts. Those are stale, so I re-measured against **`development` on the canonical repos** — resolving each config the way the workflow does, computing which spec files become dead, and fetching the newly-dead files to count the `@e2e` anchors in them.

**Three repos lose coverage credit. Every other repo loses none.**

| repo | config CI runs | spec files | newly dead | `@e2e` anchors that stop counting |
|---|---|---|---|---|
| **launchpad** | `tests/e2e/ci/playwright.config.ts` | 26 | 20 | **183** |
| **openregister** | `tests/e2e/ci/playwright.config.ts` | 63 | 30 | **174** |
| **hermiq** | `tests/e2e/spec-coverage/playwright.config.ts` | 21 | 14 | **102** |
| | | | | **459 total** |

**The other eleven repos with a second config move by zero anchors.** opencatalogi, openconnector, doriath, softwarecatalog, larpingapp, decidesk, pipelinq, procest, portaliq, scholiq, docudesk, openbuild all carry `tests/e2e/playwright.config.ts` with `testDir: __dirname` — which resolves to `tests/e2e`, the same scope the root config had. They shed 0–2 files each (`visual/**`, `docs-screenshots.spec.ts`, whose sub-config drops the projects that pulled them back in) and **those files carry no anchors**. Root-only repos (nldesign, shillinq, petstore) are untouched.

Two parser shapes this exercised, both handled without a special case:

- **`testDir: __dirname`** — an identifier, not a string literal. No string is parsed, so the fallback applies: an absent/unparsable `testDir` means *the config's own directory*, which for `tests/e2e/playwright.config.ts` **is** `tests/e2e`. Correct by construction.
- **`testIgnore: IGNORED`** (larpingapp) — also an identifier. `_patterns()` returns `None`, so nothing is treated as ignored. That is the conservative direction: fewer findings, never an invented one.

Scenario-level delta, measured on the two repos I hold checkouts of: **openregister 796 → 860 (+64)**, **hermiq 433 → 491 (+58)**; every other local repo unchanged. Anchors and scenarios are different units — one tag can prove several scenarios and one scenario can be tagged several times — so the 459 above is the input that stops counting, not the finding count.

launchpad is confirmed live on canonical (`tests/e2e/ci` + root, both present), so **launchpad#85 and openregister#2410 have not landed the collapse yet** on `development`. This fix is correct either way; the `test_a_repo_that_COLLAPSED_to_one_root_config_still_covers` arm pins that.

PR #333, opened 2026-08-10
fix(gate-54): a polymorphic reference had no authorable form

rubenvdlinde:
## The defect

Gate 54 `relation-dialect` cannot be closed for an identifier whose **target schema varies per row**. Measured 2026-08-10 on scholiq (gate package `94c855b`):

| property | discriminator |
|---|---|
| `CoursePackageImportReport.entries[].targetId` | sibling `targetType` — Course/Lesson/Material/Item/LtiToolPlacement/Assignment/Rubric |
| `LearningRecordExport.coverageReport[].sourceId` | sibling `sourceSchema` |

Both arms fail, exactly as the cross-app case did before `x-external-register` existed:

```
WITHOUT $ref -> (b) "relation-shaped property ... lacks canonical $ref (ADR-062 rule 7)"
WITH    $ref -> resolves, and is a LIE: OpenRegister would resolve EVERY row against
                the one schema the author happened to pick, including rows that are
                a different schema.
```

The only remaining route to green was rewording the description until `_RELATION_DESC_RE` stopped matching — degrading documentation to dodge a regex, which **this module's own docstring already names as the thing a gate must never reward**.

## The fix

`x-relation-schema-field: <siblingPropertyName>`, mirroring the `x-external-register` precedent. **It is not a waiver** — it is a claim the gate verifies:

- the named sibling **must exist** on the same object. Without this, `x-relation-schema-field: anything` would be a blanket silencer.
- marker **+ `$ref`** is still reported — the two claims contradict and the `$ref` is the half OpenRegister acts on.
- it suppresses **only** the two `$ref` rules. Filter placement and token validation still apply (control test).

Findings stay 1:1 with defects: a bogus discriminator does *not* also emit "lacks canonical $ref". My own new test caught that double-count before this shipped — it is the ratio defect the docstring pins for gate 53.

`_collect_properties` gains an **optional** `parents` out-dict rather than a sixth tuple element, so every existing caller and every existing test that unpacks the 4-tuple is untouched.

## Evidence — both directions

- **6 new tests**, including `test_the_true_positive_survives_without_the_marker`: removes the marker from the *same* fixture and asserts it goes red. Without that control the exemption would just be a hole.
- `test_check_relation_dialect`: **34/34**. Full discovered helper suites: **69 passed, 0 failed** (2 pre-existing quarantines untouched).
- **On the real register**: the *unmodified upstream* checker **still reports both findings against the annotated scholiq register**. The descriptions are byte-identical and still trip the regex — the app side dodged nothing. Only the gate learning the concept clears them.

## Note for reviewers

This touches shared fleet tooling consumed live at `main`, so I have **not** self-merged. ADR-062 rule 7 should gain a sentence naming the polymorphic form; happy to open that against hydra once the shape here is agreed.

Consumer: ConductionNL/scholiq (register annotation held behind this).

### DIFF
--- a/hydra-gates/scripts/lib/check_relation_dialect.py
+++ b/hydra-gates/scripts/lib/check_relation_dialect.py
@@ -50,6 +50,13 @@ lines are advisory and never fail the gate):
                         with ``format: uuid`` + a relation-shaped description
                         but NO ``$ref`` (conservative: format:uuid alone is
                         NOT enough — NC-user-id fields legitimately lack $ref).
+                        Two identifiers are exempt because no local schema key
+                        can express them: ``x-external-register`` (the target
+                        lives in another app's register) and
+                        ``x-relation-schema-field`` (POLYMORPHIC — the target is
+                        named per row by a sibling discriminator). Neither is a
+                        waiver: the discriminator must be a real sibling
+                        property, and both forbid a ``$ref``.
   c. FILTER PLACEMENT — ``x-relation-filter`` anywhere except directly on a
                         property; or on a property that is not a relation
                         (filter on non-relation is inert).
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
## Blast-radius sweep: provably zero

Paired this shared-tooling change with a consumer sweep before asking for a merge, since `.github@main` is consumed live by every repo's CI.

Ran the **unmodified upstream** `check_relation_dialect.py` and the **patched** one over every register in the fleet checkout and diffed the findings file-by-file:

```
registers compared: 349
differing:            0
```

Byte-identical findings on all 349. And no repo authors the new marker today:

```
grep -rl "x-relation-schema-field" */lib/Settings/  ->  none
```

So the change is purely additive: it can only ever affect a property that opts in by declaring `x-relation-schema-field`, which nothing does yet. No existing green goes red, no existing red goes green.

CI here is green (package invariants + the discovered helper suites; 69 passed, 0 failed, 2 pre-existing quarantines untouched).

Not self-merging — this targets `main` on shared fleet infra. Consumer waiting on it: ConductionNL/scholiq#313 (draft).

PR #336, opened 2026-08-10
fix(gate-47): a pure rename is not a security change

rubenvdlinde:
## The defect

`check_security_cochange.changed_lines()` asked git for the diff of **one** path:

```
git diff -U0 base...HEAD -- src/dialogs/X.vue
```

A pathspec is applied **before** rename detection runs. Naming only the destination removes the source side of the pair from the diff, so git has nothing left to match against and reports the destination as `new file mode` with **every line added**.

A pure `git mv` of a file that merely *contains* a security token (`requesttoken`, `IUserSession`, …) therefore classified as a security-touching change of the whole file — the same *classify-the-file-not-the-hunks* defect this module was written to remove, arriving through the pathspec instead of through grep.

## Measured

On **pipelinq#763**. `git mv src/components/ContactImportDialog.vue src/dialogs/` — required by gate-13 (modal-isolation) — is reported by git itself as:

```
src/{components => dialogs}/ContactImportDialog.vue | 0
1 file changed, 0 insertions(+), 0 deletions(-)
similarity index 100%
```

gate-47 read that as **230 added lines** and reported:

```
[gate-47] security-change-has-tests: FAIL — 1 security-touching change(s) without a test co-change
```

on a PR that changed no code at all. The remaining remedies were the opt-out or a token test — and this module's own docstring already names that outcome:

> A gate whose finding cannot be acted on truthfully is an unclosable gate, and unclosable gates are how a suite loses its readers.

## The fix

Read the rename map once, **unscoped** (`git diff --name-status -M`), and pass **both** spellings in the pathspec so git can pair them again.

- a pure rename yields **no** lines
- a rename **with** edits still yields exactly its real hunks

So nothing that used to fire stops firing.

## Tests (+3)

New class `RenamesAreNotContentChanges`:

| test | asserts |
|---|---|
| `test_fp_a_pure_rename_is_not_a_security_change` | a pure rename of a CSRF-carrying component is **not** reported |
| `test_tp_a_rename_that_also_edits_security_code_still_fires` | a rename that **also** adds a CSRF header **is** still reported |
| `test_the_rename_map_pairs_source_to_destination` | the map pairs destination → source |

## A note on the fixture — it nearly shipped broken

`git mv` **fails when the destination directory does not exist**, and `_Repo._git` swallows the error (`check=False`, output captured). All three tests initially moved into a directory that did not exist, so they changed nothing, committed nothing, and diffed an **empty change set** — and `assertEqual(security, [])` passes for the wrong reason on an empty diff.

Only the rename-map test caught it, because it asserts a **non-empty** result and therefore *cannot* pass on an empty diff.

`_Repo` gains a `move()` helper that creates the destination directory and **raises** on a failed `git mv`.

## Both directions proven

| arm | result |
|---|---|
| repaired fixture + **original** gate | FP test **fails** (`['src/dialogs/ImportDialog.vue'] != []`), rename-map test **errors** |
| repaired fixture + **fixed** gate | **23 tests OK** |

The pre-existing 20 tests are untouched and still pass.

## Blast radius

pipelinq consumes this workflow and the gate package at `@main`, so this reaches every repo in the fleet as soon as it lands. The change is confined to how the diff is *read*; no token vocabulary, path pattern, or opt-out behaviour is altered.

### DIFF
--- a/hydra-gates/scripts/lib/check_security_cochange.py
+++ b/hydra-gates/scripts/lib/check_security_cochange.py
@@ -170,23 +170,44 @@ def line_is_security_relevant(line: str) -> bool:
     return bool(_CODE_TOKEN_RE.search(line))
 
 
-def changed_lines(base_ref: str, path: str, cwd: str) -> list[str]:
+def changed_lines(base_ref: str, path: str, cwd: str,
+                  old_path: str | None = None) -> list[str]:
     """The added/removed lines for *path*, without diff framing.
 
     `-U0` so no context line is mistaken for a change: context is precisely
     the code the PR did NOT touch, and treating it as touched is the defect
     this function exists to remove.
+
+    *old_path* is the pre-rename spelling when the file was renamed, and it
+    MUST be in the pathspec alongside *path*. A pathspec is applied BEFORE
+    rename detection runs, so asking for the destination alone deletes the
+    source side of the pair from the diff and git has nothing left to match
+    it against — it then reports the destination as `new file mode` with
+    every line added. A pure move of a file that merely CONTAINS a security
+    token (`requesttoken`, a session lookup) therefore classified as a
+    security change of the whole file, which is the same
+    classify-the-file-not-the-hunks defect this module was written to
+    remove, arriving through the pathspec instead of through grep.
+
+    Measured 2026-08-10 on pipelinq#763: `git mv src/components/X.vue
+    src/dialogs/X.vue` (`similarity index 100%`, `0 insertions(+), 0
+    deletions(-)`) was reported by this gate as 230 added lines and one
+    security-touching change.
+
+    With both spellings present git pairs them again, so a pure rename
+    yields no lines and a rename-with-edits yields exactly its real hunks.
     """
+    pathspec = [path] if old_path is None else [old_path, path]
     proc = subprocess.run(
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
**This is now blocking a repository, not just a PR.**

pipelinq#763 (gate-13 modal-isolation) merged into `development`. It contains the `git mv` that triggers this false positive. On the **pull_request** run the gate-47 opt-out in the PR body cleared it, and the run was fully green — but a **push** event has no PR body, so the `development` push run [31436645687](https://github.com/ConductionNL/pipelinq/actions/runs/31436645687) is red on exactly one gate:

```
[gate-47] security-change-has-tests: FAIL — 1 security-touching change(s) without a test co-change
[hydra-gates] RESULT: 1 GATE(S) FAILED.
```

on a file whose diff is `1 file changed, 0 insertions(+), 0 deletions(-)`. The standing `development → beta` PR run fails for the same reason.

There is no pipelinq-side fix that is not a waiver: the opt-out cannot be expressed for a push event, and the offending file is a pure rename that gate-13 *requires*. Merging this PR turns pipelinq's `development` green with no further change in that repo, since it consumes the gate package at `@main`.

Checks here are green (15 pass / 15 skipping), including `Package invariants`, which runs the +3 tests — and those tests fail against the pre-fix gate, so they are load-bearing rather than decorative.

Issue #347, opened 2026-08-11
gate-61 listener-work-placement is structurally dead on every --full / workflow_dispatch run (hides 3 real findings in openregister)


rubenvdlinde:
## Summary

`gate-61 listener-work-placement` is **structurally incapable of reporting a result on any `--full` / `workflow_dispatch` run**. It reports `NOT APPLICABLE`, which does not count against coverage, so it is invisible in the `COVERAGE:` line as well. On openregister this hides **3 real findings**.

This is the falsely-NOT-APPLICABLE dead-gate shape: the exemption reason names a scope condition that does not hold on the run that printed it.

## Observed

openregister full-scope run [31488412842](https://github.com/ConductionNL/openregister/actions/runs/31488412842), Hydra Gates job `93770857036`, gate package `b8c7eade` (@main). The same log contains both of these lines:

```
[hydra-gates] Base ref: n/a — --full requested, scanning the entire tree.
...
[gate-61] listener-work-placement: NOT APPLICABLE — the diff against 'origin/development'
put every post-event registration out of scope, so NONE were inspected.
```

There was no diff. The gate excluded itself by naming one.

## Root cause (read in the source, not inferred)

1. `bin/hydra-gates` — on `--full` (`SCOPE_DIFF=0`) it builds the runner argv as `set -- "${APP_DIR}"` and only prepends `--scope-to-diff --base "${BASE_REF}"` when `SCOPE_DIFF=1`. **The runner never receives `--base`.**
2. `scripts/run-hydra-gates.sh:204` — `BASE_REF` therefore keeps its hardcoded default `"origin/development"`.
3. `scripts/run-hydra-gates.sh:7854` — gate-61 calls `check_listener_placement.py . --base "${BASE_REF}"` **unconditionally**. The in-file comment says this is deliberate ("this gate is always diff-scoped in both modes … Switching an unscoped run to `--all` was tried and reverted here: the builder runs unscoped, so it would have surfaced the whole backlog as blocking findings on every build").
4. On a full-repo run of `development`, `origin/development...HEAD` is empty → every registration falls out of scope → the checker exits 3 (`EXIT_EMPTY_SCOPE`) → the runner calls `_skip 61 … na`.

The revert referenced in (3) was correct for the case it addressed. The defect is that it treats **two different unscoped modes as one**: the builder's implicit unscoped run, and an explicit whole-tree audit that a human asked for by pressing "Run workflow".

## Positive control — the gate is blind, not clean

[...]

PR #349, opened 2026-08-11
fix(gate-24): the JS probe knew one of two registration APIs, so it cleared repos that have leaves

rubenvdlinde:
gate-24 (integration-parity) decides applicability from its own subject matter:
does this repo register integration leaves at all? The PHP half looks for
`new LeafDescriptor(` in lib/. The JS half looked only for
`registerIntegration(` in src/.

There are TWO canonical JS registration forms, and the gate knew one.
`registerIntegration(descriptor)` is the convenience wrapper exported from
@conduction/nextcloud-vue; the registry object it wraps is called directly as
`OCA.OpenRegister.integrations.register(descriptor)`, which is equally canonical
— it is how nextcloud-vue registers its OWN built-in leaves
(src/integrations/builtin/files.js).

So an app using the direct form was told, verbatim, "this repo registers no
integration leaves at all". That sentence is not cosmetic: it routes the gate to
`na`, which does NOT count against coverage, instead of `structural`, which
does. The one gate that exists to correlate the server and JS halves of a leaf
therefore switched itself off precisely in repos that have halves to correlate,
and did so while printing a confident absence claim.

MEASURED, `--full`, on ConductionNL/decidesk at e03f5932:

  [gate-24] integration-parity: NOT APPLICABLE — ... this repo registers no
  integration leaves at all — no `new LeafDescriptor(` in lib/ and no
  `registerIntegration(` in src/.

decidesk registers exactly one leaf, `decidesk-decisions`, at
src/integrations/registerDecisionsLeaf.js:175 via
`target.OCA.OpenRegister.integrations.register(decisionsLeafDescriptor)` — and
that id was present in the LIVE JS registry in the same repo's Playwright run,
read out of `window.OCA.OpenRegister.integrations.list()`. The gate said the set
was empty while the runtime enumerated a member of it.

POSITIVE CONTROL, old probe vs new, over the fleet checkout:

  decidesk       old=0 new=1   <- false negative, fixed
  openregister   old=0 new=1   <- false negative, fixed
  openconnector  old=1 new=1   <- no regression
  procest        old=1 new=1   <- no regression
  hermiq         old=1 new=1   <- no regression
  nldesign       old=0 new=0   <- correctly negative
  docudesk       old=0 new=0   <- correctly negative

and on decidesk the verdict moves from a false `na` to `SKIPPED (structural)`,
which moves the coverage line from "58 of 58 applicable gates ran" to "58 of 59"
— the gap becomes visible instead of being absorbed.

`integrations\.register` cannot collide with `unregister(`, because the
qualifier is part of the match, nor with `registerIntegrationIcons(`, because
`\(` anchors the end.

Why this matters beyond the count: the drift gate-24 exists to catch is live
right now. openregister advertises `kvk` and `opencorporates` over OCS
capabilities while @conduction/nextcloud-vue never registered either in JS
(ConductionNL/nextcloud-vue#630), and it went unnoticed until an app's e2e test
compared the two lists by hand. A parity gate that cannot tell it has leaves is
a parity gate that cannot find that.

NOT FOR IMMEDIATE MERGE. This turns a silent `na` into a counted `structural`
gap in at least two repos, so it can redden a `--require-full-coverage` run
mid-wave — the same reason #347 is being held. Land it once the wave settles.

---

### DIFF
--- a/hydra-gates/scripts/run-hydra-gates.sh
+++ b/hydra-gates/scripts/run-hydra-gates.sh
@@ -3840,12 +3840,39 @@ _parity_has_js=0
 if [ -d lib ] && grep -rqE 'new[[:space:]]+LeafDescriptor[[:space:]]*\(' lib/ --include='*.php' 2>/dev/null; then
     _parity_has_php=1
 fi
-if [ -d src ] && grep -rqE '\bregisterIntegration[[:space:]]*\(' src/ 2>/dev/null; then
+# THE JS PROBE MUST MATCH BOTH SUPPORTED REGISTRATION APIs, NOT ONE.
+#
+# `registerIntegration(descriptor)` is the convenience wrapper exported from
+# @conduction/nextcloud-vue (src/integrations/registry.js). The registry object
+# it wraps is equally canonical and is called directly as
+# `OCA.OpenRegister.integrations.register(descriptor)` — including by
+# nextcloud-vue's OWN built-in leaves (src/integrations/builtin/files.js).
+#
+# Probing for only the wrapper made this gate produce a FALSE ABSENCE CLAIM, and
+# the claim was load-bearing: an app registering leaves via the direct form was
+# told "this repo registers no integration leaves at all", which routes it to
+# `na` (not counted as a gap) instead of `structural` (counted). So the one gate
+# that exists to correlate the server and JS halves of a leaf switched itself
+# off, silently, in exactly the repos that have leaves to correlate.
+#
+# MEASURED across the fleet checkout, both forms are live:
+#   registerIntegration(     -> hermiq/src/integration-leaf.js,
+#                               openconnector/src/integration.js,
+#                               procest/src/main.js
+#   integrations.register(   -> decidesk/src/integrations/registerDecisionsLeaf.js,
+#                               openregister/src/main.js
+# decidesk registers `decidesk-decisions` through the second form and was
+# reported by this gate as having no leaves at all, on a --full run, while that
+# very id was present in the live JS registry.
+#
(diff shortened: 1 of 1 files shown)


rubenvdlinde:
Superseded by #370, which is merged.

Same runner change, rebased onto `main`, plus the two things this PR was missing:

1. **A fixture.** A paired arm on the existing `gates-23-33/clean` bundle: one overlaid file registering a leaf through the direct form. `clean/` alone stays `NOT APPLICABLE` (asserted, so an over-correction that made gate-24 structural everywhere is caught); `clean/` + that file must be `SKIPPED (structural)`. Observed **3 red before, 3 green after**.

2. **A measured blast radius — and it corrects this PR's body.** This description says the verdict moves "in at least two repos" and names openregister. Running the runner over each fleet checkout, old vs new, nothing else changed:

| repo | old | new |
|---|---|---|
| **decidesk** | `NOT APPLICABLE` | **`SKIPPED (structural)`** |
| openregister | `PASS` | `PASS` |
| openconnector | `PASS` | `PASS` |
| hermiq | `PASS` | `PASS` |

openregister ships `scripts/check-integration-parity.sh`, so gate-24 takes the wrapper path and passes either way. **Exactly one repo changes**, and it is filed as ConductionNL/decidesk#466 with the one-file remedy so the redness has an owner.

Worth recording: the fixture's **first draft was a dead test** — it went green against the pre-fix runner because its own explanatory comment named the wrapper call, and gate-24's probe is a plain grep over file text. That is #358's prose-parsing shape, met while fixturing a different gate. The warning is now written into the fixture file.

Branch `fix/gate-24-detects-direct-registry-register` can be deleted.

PR #364, opened 2026-08-11
fix(gate-16): an EMPTY SCOPE is not a PASS (.github#361)

### LINKED ISSUE #361: gate-16 spec-coverage prints PASS over 1466 findings on every --full run: a hardcoded 'origin/development' base manufactures an empty diff (sibling of #347, but falsely GREEN)
Sibling of #347 (gate-61 dead on `--full`), same root cause, **worse symptom**: gate-61 emits `NOT APPLICABLE`, which the runner excludes from the verdict. **gate-16 emits `PASS`**, which counts as an applicable gate that ran and passed — so it inflates the COVERAGE line and reads as a clean bill of health.

## The line

`hydra-gates/scripts/lib/check_spec_coverage.py:508`

```python
base_ref = os.environ.get("HYDRA_GATE_BASE_REF", "origin/development")
```

The default is **hardcoded and non-empty**. On a `--full` / `workflow_dispatch` run the runner sets no base ref, so instead of scanning the tree the gate silently diffs `origin/development...HEAD`. On the `development` branch itself that diff is **empty** — zero changed files, zero findings, `PASS`.

## Positive control — one tree, one checker, only the scope input differs

Canonical package at `3c8da4c`, run against a worktree pinned to pipelinq `development` @ `206db4be`:

```
$ env -u HYDRA_GATE_BASE_REF  python3 .../check_spec_coverage.py <tree>
# count=0

$ HYDRA_GATE_BASE_REF=origin/beta python3 .../check_spec_coverage.py <tree>
...
src/views/widgets/MyLeadsWidget.vue::onViewAll — missing @spec
# count=185

$ python3 .../check_spec_coverage.py <tree> --mode report
uncovered_count = 1466
```

And CI agrees. pipelinq run [31505417691](https://github.com/ConductionNL/pipelinq/actions/runs/31505417691), whose own log says

```
[hydra-gates] Base ref: n/a — --full requested, scanning the entire tree.
```

prints

```
[gate-16] spec-coverage: PASS
```

**over 1466 untagged methods**, inside `COVERAGE: 58 of 64 declared gates reported a result (6 not applicable; 58 of 58 applicable gates ran)`.

So the same repository has three gate-16 numbers, and the reassuring one is the fiction:

| scope | result |
|---|---|
| `--full` (what a fleet audit runs) | **PASS** |
| diff vs `origin/beta` | 185 |
| `--mode report` (whole tree) | **1466** |

## It is exactly one checker

Grepping every `HYDRA_GATE_BASE_REF` consumer in `scripts/lib/`, `check_spec_coverage.py` is the **only** one with a non-empty hardcoded default. The rest default to `""` and correctly report NOT APPLICABLE with an honest reason — in the very same run, gates 29/47/48 say:

> this run is not diff-scoped (no `--scope-to-diff` / no base ref), so there is no change set to classify

## Suggested fix — gate-19 already does it correctly

`check_e2e_coverage.py` is the model: no base ref means *scan everything*, and an empty diff **when a base ref was given** is reported as `EMPTY SCOPE` (exit 3), explicitly "not a pass".

```python
base_ref = os.environ.get("HYDRA_GATE_BASE_REF")
if base_ref:
    changed = changed_lines(base_ref, app_dir)
    stripped = spec_tags_removed(base_ref, app_dir)
else:
    # --full: no diff to scope to. Audit the whole tree.
    ...
```

Two properties worth preserving from gate-19's implementation:

1. **No base ref ⇒ whole tree**, never a manufactured diff.
2. **Base ref given but diff empty ⇒ its own distinct exit**, never `PASS`. A gate that had nothing to look at must not print the same word as a gate that looked and found nothing.

⚠️ Landing this turns a permanently-green gate red in every repo at once, exactly as #347 does — so it probably wants the same release handling.

Found while re-measuring pipelinq's full-scope table after ConductionNL/pipelinq#769.

rubenvdlinde:
Closes #361.

gate-16 is diff-scoped by design (ADR-020). On a full-repo run there is no diff, so the runner's unguarded `BASE_REF` let the checker fall back to `origin/development` — which on `development` diffs the branch against itself. Result: `# count=0` → **PASS over a scope nothing was read from**.

**This is worse than the sibling #347.** gate-61's `NOT APPLICABLE` is excluded from the verdict; **gate-16's `PASS` counts toward "N of N applicable gates ran"**. It does not hide a number — it falsifies a verdict. Caught live on openregister#2422, where gates 19/25/26 printed `NOT APPLICABLE` and gate-16 printed `PASS` **four lines apart over the same empty scope**.

## Measured in six repos, one tree each, changing only this input

| repo | full (today) | explicit base | report mode |
|---|---|---|---|
| pipelinq | 0 → PASS | 185 (beta) | **1466** |
| shillinq | 0 → PASS | **808** (empty-tree) | — |
| docudesk | 0 → PASS | 290 (beta) | 471 |
| openregister | 0 → PASS | 232 (beta) | 234 |
| procest | 0 → PASS | 113 (beta) | 445 |
| larpingapp | 0 | 0 | 0 — **genuinely clean, positive-controlled** |

larpingapp matters: it proves this is not a blanket "every repo hides debt" claim.

## Why not a literal copy of gate-19's guard

gate-19's else-branch falls back to a **full sweep**. For gate-16 that would flag the entire legacy `@spec` surface — the wrong contract per ADR-020, and a **false RED in every repo**. Reporting `NOT APPLICABLE` with a reason naming the *absence* of a diff is honest and cannot go false-RED.

## Verification

- full run before: `[gate-16] spec-coverage: PASS` → after: `NOT APPLICABLE — full-repo run computed NO diff … This is NOT a pass.`
- checker positive control on a real tree (openregister): `origin/beta` → `# count=232`, `origin/development` → `# count=0`
- **the diff-scoped arm is byte-identical** to the original invocation (the diff moves the same two lines inside an `if`), so `--scope-to-diff` behaviour is unchanged by construction

⚠️ My first draft passed `scope` as the `_skip` category. Anything outside `na|structural|wiring` is treated as an **internal-error FAIL** — it would have produced exactly the false RED this change prevents. Now `na`.

Lands after #359 (the acceptance suite), which is merged.

### DIFF
--- a/hydra-gates/scripts/run-hydra-gates.sh
+++ b/hydra-gates/scripts/run-hydra-gates.sh
@@ -3051,24 +3051,30 @@ if [ -d lib ] || [ -d src ]; then
         # wrong contract per ADR-020, and a false RED in every repo. Reporting
         # NOT APPLICABLE with a reason that names the ABSENCE of a diff is both
         # honest and incapable of going false-RED.
-        if [ "${SCOPE_TO_DIFF}" = "1" ]; then
-            HYDRA_GATE_BASE_REF="${BASE_REF}" \
-                python3 "${_sc_lib_dir}/check_spec_coverage.py" . \
-                >> "${_sc_log}" 2>"${_sc_err}" || true
-        else
-            _sc_ran=0
-            # Category MUST be one of na|structural|wiring — `_skip` treats
-            # anything else as an internal-error FAIL, which would be exactly
-            # the false RED this change exists to avoid. `na` is the honest
-            # one: the gate is not applicable to a run that has no diff, and
-            # the reason names the ABSENCE of a diff rather than claiming a
-            # diff excluded something (the invariant .github#359 asserts).
-            _skip 16 "spec-coverage" na "full-repo run computed NO diff, and gate-16 is diff-scoped by design (ADR-020) — so NO changed method was inspected and @spec traceability is UNVERIFIED by this run. This is NOT a pass. Re-measure with an explicit base (HYDRA_GATE_BASE_REF=origin/beta) or run with --scope-to-diff."
-        fi
-        if [ "${_sc_ran}" != "0" ] && ! grep -qE '^# count=[0-9]+$' "${_sc_log}" 2>/dev/null; then
+        # ALWAYS run the checker, on every scope. An earlier draft of this fix
+        # skipped the invocation entirely on a full run — and the package's own
+        # `test_gate_crashed_checker_is_not_a_finding.sh` caught the regression:
+        # a CRASHED checker became invisible at full scope, because nothing ran
+        # to crash. Running it costs nothing and keeps that invariant intact.
+        HYDRA_GATE_BASE_REF="${BASE_REF}" \
+            python3 "${_sc_lib_dir}/check_spec_coverage.py" . \
+            >> "${_sc_log}" 2>"${_sc_err}" || true
+        # Order matters: WIRING first (did the checker finish?), then SCOPE (is
+        # its answer meaningful?). A crash must never be reported as an empty
+        # scope, and an empty scope must never be reported as a pass.
+        if ! grep -qE '^# count=[0-9]+$' "${_sc_log}" 2>/dev/null; then
             _sc_ran=0
             _sc_why=$(head -3 "${_sc_err}" 2>/dev/null | tr '\n' ' ' | cut -c1-200)
             _skip 16 "spec-coverage" wiring "check_spec_coverage.py did not complete — it never printed its terminal '# count=' marker, so NO changed method was inspected and @spec traceability (ADR-003/ADR-020) is UNVERIFIED by this run. Checker output: ${_sc_why:-<empty>}. See ${_sc_err}."
+        elif [ "${SCOPE_TO_DIFF}" != "1" ]; then
+            # The checker RAN and finished — but on a full-repo run it had no
+            # diff to scope to, so `# count=0` means "nothing was read", not
+            # "nothing is wrong". Category MUST be one of na|structural|wiring:
+            # `_skip` treats anything else as an internal-error FAIL, which is
+            # exactly the false RED this change exists to prevent (my first
+            # draft passed `scope` and would have done that fleet-wide).
+            _sc_ran=0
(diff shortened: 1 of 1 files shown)


rubenvdlinde:
### A seventh repo for the table, and it looks like larpingapp, not like pipelinq

Measured after this merged, but with the **pre-fix** package (`@81c8c97`) so it
is directly comparable to the six rows above — openconnector, one tree
(`origin/development` @`7c1d9c6d`), changing only the base:

| repo | full (pre-fix) | `origin/beta` | `origin/main` | report mode |
|---|---|---|---|---|
| **openconnector** | **0 → PASS** | **0** | **30** | **30** |

**Positive-controlled, committed** — an untagged public method in
`lib/Service/ActionAuthService.php` *and* an untagged Vue method in
`src/components/CircuitBreakerBadge.vue`. Every arm moved by exactly **+2** and
named both plants (`count=2` / `count=32` / `uncovered_count 32`); reverted,
all four numbers returned.

Two things worth adding to the record:

1. **openconnector's `origin/beta` is 0 too**, across **467 commits** of
   divergence from `development`. So this repo shows **two consecutive zeros**
   where pipelinq/docudesk/openregister/procest all showed a large second
   number. Two zeros in a row are exactly what a dead gate looks like, and only
   the planted control tells them apart. Alongside larpingapp, that is now two
   repos where the honest answer really is "almost nothing hidden" — 30
   methods, all backend, concentrated in `SynchronizationService` (9) and
   `SynchronizationContractService` (9), **zero frontend** (and the frontend
   arm of the control fired, so that zero is real too).

2. **The spread is now 0 → 1466 across seven repos.** Anyone extrapolating
   pipelinq's ratio would overstate openconnector by roughly **fifty times**.
   Worth stating in the issue text so the next agent measures rather than
   scales.

### On "why not a literal copy of gate-19's guard"

Agreed, and the asymmetry is worth naming explicitly for whoever reads this
next. gate-19 on the same tree prints:

```
[gate-19] e2e-coverage: EMPTY SCOPE — diff-scoped against 'origin/development'
and NO spec file was touched. 85 spec file(s) exist here and NONE were
inspected: @e2e traceability (ADR-020) is UNVERIFIED by this run.
This is not a pass.
```

So the distinguishing property was never "does the checker default the env
var" — it was **"does an empty scope print PASS, or print something that
cannot be mistaken for one"**. gate-19 already had the second behaviour;
gate-16 had the first. Your `NOT APPLICABLE … This is NOT a pass.` gives
gate-16 the same property without inheriting gate-19's full-sweep fallback,
which is the right call for a gate whose legacy surface is the whole tree.

Issue #360, opened 2026-08-11
gate-7: hasPermission() and canAccess() still reported as unguarded IDOR after #353


rubenvdlinde:
Follow-up to #353, found by the gate acceptance suite (#359) rather than by an app.

## The residual false-positive class

`#353` widened `_GUARD_HELPER_NAME_RE` so the auth token no longer has to be the
**final** CamelCase segment. It did not remove the requirement that a **non-empty
segment sit BETWEEN the prefix and the token**, so **the auth token can never come
first**.

Consequently these are still reported as unguarded IDOR despite being guards:

- `hasPermission()`
- `canAccess()`

Both are conventional Nextcloud-ish predicate names, so this is not an exotic shape.

## Measured, not inferred

The gate acceptance suite in #359 exercises the regex directly against both the
pre-#353 and post-#353 forms.

⚠️ **Also worth correcting for the record:** the `#353` commit message states that
`canAccess` matched. Measured against **both** regexes, it never did. `#353` is
still a clean widening — nothing was lost and hermiq genuinely went 3 → 0 with all
of gate-7's own unit tests passing — but the message overstates its scope.

## Why this is filed rather than fixed here

Two agents are mid-flight on gate-7-adjacent work and #359 (the acceptance suite)
is open for review. Landing a second regex widening now would change a gate's
verdict across the fleet while that suite is still being reviewed. The fix belongs
**after** #359, and #359 should carry the fixture that pins it.

Suggested fixture, to live alongside the #353 one: `hasPermission` and `canAccess`
in the clean arm (must NOT be findings), with `canRender` and `hasChanges` staying
in the planted arm (must remain findings, since neither carries an auth token).

/cc the hydra-gates acceptance suite work in #359

PR #373, opened 2026-08-11
fix(gate-7): an authentication check is not an authorisation guard (#365)

### LINKED ISSUE #365: gate-7 accepts a 401 authentication check as an authorisation guard — 4 real IDORs read as PASS (fleet-wide false negative)
## Summary

`check_no_admin_idor.py`'s `_GUARD_BODY_RE` accepts `Http::STATUS_UNAUTHORIZED` as an authorisation guard. But a 401 is an **authentication** signal ("is anyone logged in?"), not an **authorisation** one ("may *this* caller touch *this* object?").

Nextcloud's idiomatic preamble for a `#[NoAdminRequired]` endpoint is exactly that check:

```php
$user = $this->userSession->getUser();
if ($user === null) {
    return new JSONResponse(['error' => 'Authentication required'], Http::STATUS_UNAUTHORIZED);
}
```

Any method that opens this way is exempt from gate-7 **for free**, however unguarded the direct object reference that follows is. Since the preamble is the house style, this is not a rare shape.

## Three-arm control on ONE file, changing only the guard shape

Canonical package `ConductionNL/.github` @ `81c8c97`, checker invoked directly on
`pipelinq/lib/Controller/LoyaltyController.php` @ `206db4be` (a copy in a scratch dir; no checkout modified).

| arm | file | gate-7 |
|---|---|---|
| **A** — unmodified controller, containing **four real IDORs** (`getAccount`, `getAccountHistory`, `getRedemptionOptions`, `initiateRedemption`) | as shipped | **0 findings** |
| **B** — plus `plantBareUnguarded(string $accountId)`: `#[NoAdminRequired]`, reads the account by the URL id, **no auth of any kind** | +14 lines | **1 finding** — `method=plantBareUnguarded rule=no-auth-guard-in-body` |
| **C** — plus `plantAuthenticatedButUnauthorized(string $accountId)`: **byte-identical body**, preceded only by the 401 preamble above | +23 lines | **0 findings** |

Arm B proves the gate is live on this exact file and this exact shape. Arm C, differing from B only by the 401 preamble, proves the preamble is what silences it.

## Root cause

`hydra-gates/scripts/lib/check_no_admin_idor.py:384-392`:

```python
_GUARD_BODY_RE = re.compile(
    r"OCSForbiddenException"
    r"|isAdmin\s*\("
    r"|->\s*(?:authorize|require|ensure)[A-Z][A-Za-z0-9_]*\s*\("
    r"|Http::STATUS_(?:UNAUTHORIZED|FORBIDDEN)"          # <-- UNAUTHORIZED
    r"|(?:statusCode:\s*|,\s*)(?:401|403)\b"             # <-- 401
    ...
)
```

## The real IDORs this masked

All four are in `pipelinq/lib/Controller/LoyaltyController.php`, all `#[NoAdminRequired]`, all take `{accountId}` straight off the URL and never scope it to the caller. `getAccount`'s own docblock (`:80`) states *"the caller MUST own the underlying klantId"* — a contract written down and never implemented. `LoyaltyAccountService::findAccountByKlantAndProgramme()` exists as the means to scope it and is never called from the controller.

The worst is not a read leak: `POST /api/loyalty/redemption/initiate/{accountId}/{optionId}` returns **201**, debits the victim's balance, and hands the **caller** the reward code. Filed against pipelinq separately.

## Blast radius

In pipelinq alone: gate-7 reports **0** over the whole controller tree, and **49 of 82** controller files contain both `#[NoAdminRequired]` and `Http::STATUS_UNAUTHORIZED`. Every `#[NoAdminRequired]` method in those files that opens with the preamble is silenced regardless of what it does next. Other apps in the fleet use the same preamble.

## Proposed fix — and why it should NOT be merged yet

Minimal and well-motivated: **drop `UNAUTHORIZED` and `401` from `_GUARD_BODY_RE`, keep `FORBIDDEN` and `403`.** A 403 genuinely is an authorisation decision; a 401 is not, by definition.

⚠️ This has the **false-RED shape**: it will turn gate-7 red in many repos simultaneously, and it is the same sequencing hazard as `.github#347`, `#356` and `#358`. **Filing, not merging.** The coordinator should sequence it, and every repo's existing "gate-7 clean" should be treated as unverified until it lands.

🔑 Related, and the reason this went unnoticed: gate-7's known problems so far (`.github#353`) were all **false positives** — a guard the regex failed to recognise. This is the opposite and worse shape: a **non-guard the regex accepts**. A gate whose known failure mode is "too noisy" gets its silences trusted.

rubenvdlinde:
Closes #365.

> Note on wording: this description deliberately avoids reproducing verdict-shaped
> lines. The workflow exports the PR body into the gates log, so a pasted verdict
> line reappears there and reads as a finding of the very run meant to prove the
> fix. Counts below are given in prose and in a table instead.

## What was wrong

gate-7 reported zero in **all eighteen** fleet apps. Every one of those zeros came
from a checker that accepts *"is anyone logged in?"* as an answer to *"may **this**
caller touch **this** object?"*.

Reproduced before touching anything, with a three-arm committed-plant control —
same file, same commit, same run, byte-identical data-access bodies, only the
preamble differing:

| arm | canonical gate-7 |
|---|---|
| bare unguarded method | **1** |
| the same body, preceded only by a `no user -> 401` preamble | **0** |
| the same body, plus a real per-object ownership check | 0 *(correct)* |

Scale, measured against the canonical package at `57bcb2b`: zero reported
fleet-wide, while **453 of 791** controller files carry that preamble.

**Positive-controlled in all 18 apps.** A zero from a gate looks exactly like an
empty scope, so a bare `#[NoAdminRequired]` IDOR was planted into each extracted
tree and the canonical checker re-run: it moved from 0 to 1 in every single one,
then every plant was removed and verified absent. Each zero is a measurement.

Under `#[NoAdminRequired]` Nextcloud's middleware has already rejected the
anonymous caller before the method runs, so that preamble cannot even fail. It
was nonetheless silencing the gate, through three separate guard patterns.

## Why the fix is not "drop 401 from the regex"

That is what #365 proposes, and it is wrong in **both** directions. Both are now
pinned by tests.

**Too wide.** A genuine per-object decision that answers with the wrong status —

```php
if ($account['ownerId'] !== $user->getUID()) {
    return new JSONResponse([], Http::STATUS_UNAUTHORIZED);
}
```

— is a real guard. Deleting the token turns it into a false positive, and false
positives (#353, #360) are precisely how gate-7 lost its credibility and how its
silences came to be believed.

**Too narrow.** The same authentication clause answering `403` keeps clearing.
One token of edit, made by someone chasing a green cell, buys the silence back by
making the response *worse*.

So this treats **authentication-ness as a property of the condition, not of the
status code.** An `if` whose condition tests only whether a caller identity is
*absent*, and whose consequent refuses, is blanked to same-length whitespace
before any guard pattern runs. Byte offsets, spans and reported line numbers are
unchanged; only the text the guard patterns get to see is.

Three controls keep the blanking narrow, each closing a way it could have eaten a
real guard:

1. **Polarity** — absence tests only. `if ($user !== null) { ...whole body... }`
   is a wrapper, and blanking it would erase every guard inside.
2. **Argument-free operands only** — `$this->access->canAccess($id, $user)` and
   `$account['ownerId']` can never be read as an identity, however they are
   spelled. The first is a guard; the second is object data. This one was caught
   by the existing suite, which went red on the first draft.
3. **The consequent must refuse.** A conditional that merely computes is left
   alone.

Identity recognition is deliberately spelling-agnostic. #365's own re-audit built
the opposite mistake into its triage tool — three enumerated spellings of "who is
the caller", `sessionUserId()` matching none — and manufactured 19 false positives
from that gap.

## Two companion patterns, and why they ship in the same PR

Removing the clear is what *exposes* them. Shipping without them would have
reported 45 findings in doriath alone, an app whose real gate-7 exposure was
hand-verified as zero.

- **Pattern 6 — session-identity hand-off.** Every call that receives a
  caller-supplied value must *also* receive a session-derived identity, so the
  object reference is resolved under a scope the caller cannot forge. An
  ALL-quantifier, not ANY: one unscoped data call still reports, so an audit-log
  line cannot clear a real IDOR. Suppresses **86** fleet-wide.
- **Pattern 7 — in-body ownership comparison**, whatever status it answers with,
  including the 404 chosen deliberately so a 403 cannot become an existence
  oracle. That sentence was already printed by this gate's own failure guidance;
  it was advice to a human and is now code. Suppresses **2** fleet-wide, and both
  are endpoints an earlier audit named specifically as over-reports.

## Blast radius

Canonical `development` clones from GitHub, not local checkouts — three local app
directories turned out to have no `.git` at all and silently resolved the parent
repository's ref.

| app | ref | ctrl files | files with the preamble | today | with this PR |
|---|---|---:|---:|---:|---:|
| zaakafhandelapp | `bef3a7c` | 25 | 23 | 0 | **86** |
| pipelinq | `ccf0ff3` | 82 | 49 | 0 | **52** |
| shillinq | `5d31898` | 83 | 66 | 0 | **40** |
| procest | `347f2a4` | 91 | 49 | 0 | **32** |
| hermiq | `7cd256c` | 43 | 28 | 0 | **20** |
| docudesk | `67d5201` | 32 | 25 | 0 | **19** |
| decidesk | `445776b` | 40 | 18 | 0 | **17** |
| softwarecatalog | `ea4dc7e` | 21 | 15 | 0 | **14** |
| openregister | `df815c3` | 139 | 42 | 0 | **10** |
| opencatalogi | `f6e86ab` | 24 | 7 | 0 | **9** |
| openbuild | `6772ff6` | 21 | 18 | 0 | **9** |
| doriath | `bfd6da6` | 49 | 31 | 0 | **5** |
| scholiq | `d2a0a1c` | 22 | 17 | 0 | **4** |
| launchpad | `19e9b49` | 45 | 29 | 0 | **3** |
| larpingapp | `9704abd` | 6 | 3 | 0 | **3** |
| openconnector | `6da5dfb` | 46 | 32 | 0 | **2** |
| portaliq | `65cafbd` | 10 | 1 | 0 | **2** |
| nldesign | `8e49697` | 12 | 0 | 0 | **0** |
| **total** | | **791** | **453** | **0** | **327** |

**This is a candidate count, not an exposure count.** An earlier hand-audit of a
similar candidate list found per-app precision ranging from 95% to 0%. Read the
findings; do not quote 327 as a defect count.

Spot-checked that the fix does not swallow the real ones: shillinq's
arbitrary-file-read download endpoint, its SEPA mandate export, pipelinq's
semantic-handoff routes, docudesk's cross-register preview, decidesk's motion
history, opencatalogi's unscoped search, and zaakafhandelapp's object routes are
all still reported. procest's and openconnector's counts are low because both
landed real guards earlier today; the guards were read and confirmed rather than
assumed.

## Coverage added

A new `authn-vs-authz` bundle in the gate-acceptance matrix drives the control
through the real wrapper. **Both arms keep the 401 preamble**, so the clean arm
pins that the preamble is *ignored*, not *punished* — without that, "flag anything
containing a 401" would pass. Its `expect.conf` requires the finding to name
`method=preamble` rather than the file path, because the bare arm names the file
under the old checker too; a file-level assertion would have been satisfied by a
checker with the defect intact.

**Proven able to fail:** running that bundle against the pre-fix checker turns it
red, on exactly that assertion, with the driver reporting that the gate failed for
some other reason and the fixture therefore proves nothing.

Four existing unit tests encoded the defect as correct behaviour and are inverted,
each carrying the reason. The sharpest is the decidesk `citizenAction()` case: it
sits beside `staffAction()` in one collaborator, and only `staffAction()` is an
authorisation guard — the 401 arm was borrowing credibility from the 403 arm next
door.

## Verification

- 113 unit tests for this checker, green, up from 95.
- 74 helper suites green, 2 quarantined as documented.
- 61 entry-point tests green.
- One helper suite is red — the gates 45-to-55 acceptance suite — and it is red
  identically on an unmodified canonical clone: `ajv` is not resolvable in this
  environment and the suite refuses to run rather than emit a false verdict. Not
  caused by this change.

## Not verified

- No HTTP probe against a running instance. Every claim about a specific app's
  endpoints here rests on reading source.
- The residual Pattern 6 admits: a callee that accepts the identity argument and
  ignores it. This gate sees one method body by contract, so that residual is
  symmetrical with the guidance already printed with its findings, and the
  ALL-quantifier narrows it rather than removing it.

## Merge

**Deliberately not merged.** It turns a green cell red in all eighteen apps at
once, and the sequencing is Ruben's call.

---

# Extension, 2026-08-12 — gate-7's SECOND blinding exemption: `#[PublicPage]`

> Same note on wording: prose and tables, no verdict-shaped lines.

## What was wrong

`#[PublicPage]` was skipped unconditionally, commented *"intentionally open to
unauthenticated callers"*. That is a claim about the **caller**, where the gate's
question is about the **object**. The exemption carried **no stated reason**, so
there was nothing to audit.

Reproduced first, on a nine-arm rig — one file, one commit, one run, predictions
written before the run. A byte-identical IDOR plant carrying `#[PublicPage]`,
including an **unauthenticated write to an arbitrary `$id`**, cleared the gate.

**It is not safe by construction.** opencatalogi#856, merged as `963f832a`, was
exactly this shape: `GET /api/themes/{id}` was `@PublicPage` and its `find()`
carried no register and no schema, so it resolved the caller's identifier in any
register on the instance and served an anonymous caller a municipal `zaak` status
record.

## 🔑 The exemption is the smaller half of the blinding

Deleting it and changing nothing else moves **one arm in nine**. A
`#[PublicPage]` method carrying no `#[NoAdminRequired]` never reaches the
exemption — the scope filter drops it one branch earlier. Across the fleet that
is **267 of 357 public controller methods (74 %)**, including **all 78 of
procest's**. So the scope filter widens too.

## The predicate

There is no current user, so ownership guards do not transfer. A public handler
must **constrain the lookup to a scope it declares public** rather than resolve
the identifier globally. In scope: a caller-supplied *scalar* reaches a selection
call — `array $data` is a payload (portaliq's public forms are untouched), and a
`$token`/`$secret` is the authorisation, not a selector. Cleared when every such
value reaches a constrained selection, by five spellings each read out of a fleet
app: a non-caller argument that carries a value (a bare literal such as
`_rbac: true` does **not** constrain — `963f832a`'s own message says why); a
callee declaring publicness as a complete CamelCase segment, so
`getPublicationById` is not mistaken for one; a receiver already scoped as state;
a container resolution whose result constrains a later lookup and whose miss is
refused; and one resolved delegation hop. Judged **per selector, not per call**,
because "verify in scope, then act" is the commonest correct public shape.

## Two-directional control, on a real endpoint

opencatalogi `ThemesController::show` — same file, same method, same checker,
the shipped security fix in between — is reported at `f6e86ab` and is silent at
`963f832`. `PublicationsController::show`/`uses`/`used`/`index`, fixed by the
same commit, also go silent.

## Also here, and load-bearing

A same-class helper counted as guard-bearing if its body matched a bare `throw`,
matched against source **with comments in it**. hermiq's `loadOwnedAgent()`
contains no `throw` statement at all — the word appears only in a comment
explaining why it *catches* one — and that cleared all four routed methods
calling it. opencatalogi's `getObjectService()`, a service locator that throws
when OpenRegister is absent, cleared every caller in the app, which is why the
new pattern could not see the endpoint that motivates it. The body test now reads
comment-free text, a `throw` must name an authorisation exception, and an
**ownership comparison answered with `null`** is recognised instead — so no true
clear is lost. hermiq is unmoved; so is doriath.

`TemplateResponse` no longer clears a `@PublicPage` method: its stated reason is
that Nextcloud guarantees a session, and `@PublicPage` is the annotation that
turns that guarantee off.

## Blast radius

18 canonical clones on current `development`, identical file lists. The fleet
total moves by **eight**: six under the new rule name
`publicpage-unscoped-object-lookup` (openregister 3, pipelinq 1, portaliq 1,
softwarecatalog 1) and two from the helper-evidence repair (opencatalogi 1,
launchpad 1). doriath — the app that gets scoping right — is unmoved, and so is
hermiq.

**All eight were hand-read.** One is a live exposure: openregister
`NamesController::show` resolves any identifier anonymously and returns its name,
which is #856's shape one field narrower. One is a real pre-existing `#365`-class
finding surfaced by the helper repair (opencatalogi `ListingsController::index`).
The other six are endpoints whose publicness is a **claim** — a public catalogue
keyed by a public identifier. The gate cannot tell those from an object store
without knowing which schemas are public, and that is the point: the remedy is
`@no-admin-idor-exempt <reason>`, which turns a reasonless blanket exemption into
an auditable one.

## Coverage

Unit tests 113 → 130; **nine of the seventeen new ones are red against the
pre-fix checker**. New repo-shaped acceptance bundle `publicpage-scope` under
`#359`'s matrix — both arms keep `#[PublicPage]` on every method, so the clean
arm pins that the annotation is *ignored*, not *punished*, and the subject is a
method name rather than a file path. Verified able to fail: run against the
pre-fix checker, the bundle goes red on exactly that assertion.

## Not verified

No HTTP probe against a running instance — every app verdict rests on reading
source. Pattern 2's premise (that OpenRegister's RBAC decides for an anonymous
caller) is contradicted in prose by `963f832a`'s own commit message and remains
unaudited; withholding it from `@PublicPage` methods changes nothing on
openregister today, so it is not what masks those endpoints. Two pre-existing
weaknesses are named but **not** fixed here: `scan_file` still matches its guard
patterns against comment-bearing text, and an exception-to-status mapping inside
a `catch` block still reads as an authorisation decision.

### DIFF
--- a/hydra-gates/scripts/lib/check_no_admin_idor.py
+++ b/hydra-gates/scripts/lib/check_no_admin_idor.py
@@ -11,7 +11,10 @@ Recognised guard patterns (body must contain at least one):
   - ``OCSForbiddenException`` thrown
   - ``isAdmin(`` check
   - ``->authorize*(`` / ``->require*(`` / ``->ensure*(`` service call
-  - ``#[PublicPage]`` or ``@PublicPage`` annotation on the same method
+  - ``#[PublicPage]`` / ``@PublicPage`` — but only when no caller-supplied
+    identifier reaches a lookup the caller can steer.  "May be called
+    without a session" is not "every object it can reach is public"; see
+    the Pattern 8 commentary.
   - ``Http::STATUS_UNAUTHORIZED`` or ``Http::STATUS_FORBIDDEN``
   - ``TemplateResponse`` return — SPA renderers; NC middleware already
     guarantees an authenticated session so there is no object access to gate
@@ -419,6 +422,12 @@ _GUARD_BODY_RE = re.compile(
     r"|TemplateResponse",
 )
 
+# The same set WITHOUT the ``TemplateResponse`` alternative, for `@PublicPage`
+# methods. See the note at its use in ``scan_file``.
+_PUBLIC_TEMPLATE_STRIPPED_RE = re.compile(
+    _GUARD_BODY_RE.pattern.replace(r"|TemplateResponse", ""),
+)
+
 # Patterns that indicate a CORS-headers-only body (exemption 3).
 _CORS_HEADER_RE = re.compile(r"Access-Control-Allow|applyCorsHeaders")
 # Patterns that indicate data access (disqualifies exemption 3).
@@ -498,7 +507,9 @@ _HELPER_GUARD_BODY_RE = re.compile(
     r"OCSForbiddenException"
     r"|NotPermittedException"
     r"|ForbiddenException"
-    r"|\bthrow\b"
+    r"|\bthrow\b[^;\n]{0,120}?"
+    r"(?:Forbidden|Unauthori[sz]ed|NotPermitted|NotAllowed|AccessDenied"
+    r"|Permission|NotAuthenticated|Authori[sz]ation|Tenan|Owner)"
     r"|isAdmin\s*\("
     r"|isCurrentUserAdmin\s*\("
     r"|->\s*(?:authorize|require|ensure|check|assert|guard)[A-Z][A-Za-z0-9_]*\s*\("
@@ -1111,8 +1122,35 @@ def _has_session_identity_handoff(body: str, params) -> bool:
 # authentication shape this same change just finished demoting.
 
 
-def _has_ownership_comparison_guard(cleaned: str, src: str, start: int, end: int) -> bool:
-    """True when the body compares a caller identity against object data and refuses."""
+# A consequent that DENIES, as opposed to one that merely returns.
+#
+# `_REFUSAL_RE` accepts any `return`, which is right inside a routed method —
+# there is nothing to return but the answer. It is WRONG inside a predicate
+# HELPER, where `return true` is a GRANT written in the same shape. The
+# package's own abuse control caught this: `canRenderAgent()` doing
+# `if ($agent->getOwner() === $userId) { return true; }` is a tokenless
+# predicate the fleet has decided must NOT clear the gate (`#353`/`#360`
+# coordinator note: "Modify"/"Render" are not auth tokens), and reading it as
+# an ownership guard silently reversed that decision.
+_DENIAL_CONSEQUENT_RE = re.compile(
+    r"\bthrow\b"
+    r"|return\s+null\b"
+    r"|return\s*;"
+    r"|return\s+(?:false|\[\s*\])\s*;"
+    r"|Http::STATUS_"
+    r"|(?:statusCode:\s*|,\s*)(?:401|403|404)\b"
+    r"|Response\s*\(",
+)
+
+
+def _has_ownership_comparison_guard(cleaned: str, src: str, start: int, end: int,
+                                    require_denial: bool = False) -> bool:
+    """True when the body compares a caller identity against object data and refuses.
+
+    *require_denial* tightens ``refuses`` from "any return" to an actual
+    denial — used when the body being read is a HELPER rather than a routed
+    method. See ``_DENIAL_CONSEQUENT_RE``.
+    """
     for m in re.finditer(r"\bif\s*\(", cleaned[start:end]):
         at = start + m.start()
         open_paren = cleaned.find("(", at)
@@ -1122,7 +1160,10 @@ def _has_ownership_comparison_guard(cleaned: str, src: str, start: int, end: int
         stop = _consequent_end(cleaned, close + 1)
         if stop == -1 or stop > end:
             continue
-        if not _REFUSAL_RE.search(cleaned[close + 1:stop]):
+        consequent = cleaned[close + 1:stop]
+        if not _REFUSAL_RE.search(consequent):
+            continue
+        if require_denial and not _DENIAL_CONSEQUENT_RE.search(consequent):
             continue
         cond = src[open_paren + 1:close]
         atoms: list = []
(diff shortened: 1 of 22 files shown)


rubenvdlinde:
CI settled: 16 pass, 15 skipping, none failing, none pending — count stable at 31 across eleven consecutive polls, since a check list read too early is trivially all-non-pending.

The job that matters is `Package invariants`, which runs the helper suites: seventy-five green, two quarantined as documented, none red. `test_gate_acceptance_matrix.sh` is among them, so the new authn-vs-authz bundle is enforced from here on.

One correction to the description: the gates 45-to-55 acceptance suite, reported above as red locally, is **green in CI**. It refuses to run when ajv is unresolvable rather than emit a verdict it cannot support, so the local red was the machine and not the change — and CI is what establishes that rather than my having said so.

Still deliberately not merged.

PR #403, opened 2026-08-12
fix(gates 7, 57): identity must be intrinsic, and a call that resolves nothing is not an unscoped lookup (#398)

### LINKED ISSUE #398: gate-57's foundation guard keys on the checkout directory name, so it never engages in CI (and gate-7's $this-> sanitiser false positives)
## gate-57's foundation guard is keyed on the checkout's directory name — so it has never engaged in CI, for any repo

`quality.yml` checks out with `path: app`. gate-57's hydra#106 fail-safe identifies the foundation repo by **directory name**, so under CI the directory is always `app` and the guard **never matches**.

Four-arm control, same openregister tree, only the directory name varying:

| directory | gate-57 findings |
|---|---|
| `openregister` | **0** |
| renamed | **8** |

So every gate-57 figure ever taken from a CI run is a measurement of the guard not engaging, and every figure taken from a local clone named after its repo is a measurement of it engaging. **The two are not comparable, and neither is labelled.**

Corrected figures for the current fleet-impact table: **openregister 8**, **openconnector 3**.

🔑 A guard keyed on a path component is keyed on something the caller controls and the checker cannot see. The identity should come from something intrinsic — `appinfo/info.xml`'s app id, or the repository name from the environment — not from where the tree happens to sit.

## Related, and more urgent for anyone about to act on gate-7 findings

**gate-7 fires on any correctly-scoped method that sanitises a parameter through a `$this->` helper.** Isolated in both directions:

- adding `$safeKey = $this->sanitizeKey($key)` **flips a clean method to FINDING**
- adding a session identity to that helper call **clears it**

This is a systematic false-positive class, not a per-app accident. **Adding an authorisation guard would be the wrong repair** — the method is already correctly scoped, and a guard would be dead code written to satisfy a misreading.

⚠️ Anyone triaging the **+411 fleet-wide gate findings** from the `#373`/`#376`/`#377`/`#378`/`#379`/`#380` wave must be briefed on this before touching gate-7. **The size of the class is currently unquantified** — a first screening attempt was invalid (it matched the 401 preamble itself) and was discarded rather than reported.

## Context: what the wave actually did to the fleet

Measured OLD `487e8d23` vs NEW `97165ed4`, both `--full`, no delta base, each app's current `development`:

**6205 → 6616 findings, +411, with zero application-code change.**

Every app sees this on its **next run** — all 18 callers use `quality.yml@main` with `enable-hydra-gates: true` and **no `hydra-gates-ref` pin**. There is no release to wait for.

Nearly all of the +411 is pre-existing debt now visible rather than anything new. Verified on a real case: in docudesk's *same file*, `batchUpload()` (passes a session userId) is correctly suppressed while `folderBatch()` (caller-supplied `folderId`, only a 401 preamble) is reported. That is the gate working.

**The two exceptions are the two above.**

rubenvdlinde:
Closes #398.

Two defects the fleet-impact measurement exposed. Both reproduced with a positive control before anything was changed, both fixed with arms proving the fix in each direction.

Note on the numbers below: they are written as prose deliberately. A PR body is exported into the gates job environment and printed into the log, so pasting verdict-shaped lines plants them inside the artefact meant to prove the gates are clean.

## gate-57 identified the foundation repo by its checkout directory name

The hydra-gates job checks the app out with a fixed path of `app`, so under CI the directory basename is that same literal string for every repository in the fleet. The hydra#106 fail-safe compared that basename against a set of two app names, so it has never engaged in CI, for any repo, and nothing in the log said so.

Four-arm control on one openregister tree, varying only the directory name: named after the repo it reports nothing and prints a skip reason, renamed it reports eight. A leaf app reports three under both names, which is what makes this a test rather than an observation.

Identity now comes from `appinfo/info.xml`'s id element. It is intrinsic, so it reads the same under the CI path, in a worktree and in a fleet sweep. I deliberately did not use the repository name from the environment: `main()` groups its arguments by app root and judges each root separately, so a process-wide value would stamp one repository's identity onto every root, which is the same defect in different clothes. When info.xml is absent the basename remains the fallback, and every run now prints the resolved id, the source it came from and whether the guard applies. A fail-safe that silently stops being one is the defect being removed, so the fallback names itself.

Fleet effect, and the sign is the opposite of what the issue predicted: the corrected figures of eight and three describe what CI has been producing all along, unlabelled. Landing this takes those two repos to zero with a stated reason and leaves every other app untouched. That zero is a skip, not a clean bill of health, and the finding notes say so.

## gate-7 fired on correctly-scoped methods that sanitise through a helper

Pattern 6's second clause vetoed on every call receiving a caller-supplied value, whether or not that call could resolve anything. Adding a sanitiser call to a clean method flipped it to a finding while the data access beside it was unchanged and still carried the session identity. Adding an authorisation guard would have been the wrong repair.

The fix is two halves and both are required. Taint propagation follows the caller's value into the locals it is laundered into. Only then does a call that provably resolves nothing stop vetoing: a same-class or resolved-collaborator helper whose body is read and found free of data access, mutation and request reads, plus PSR-3 logging. Taint stops at a value produced by a call that also received the session identity.

Sizing was done on the checker's own decision path, not by looking for a session identity in the body, which is the screen that was discarded for matching the very preamble it exists to discount. The instrument was positive-controlled to be byte-identical to the checker on a real app first. Of the findings across the fifteen core apps with a tracked controller directory, a hundred and thirty-two were second-clause vetoes and forty-one had the isolated shape. All forty-one bodies were read by hand, splitting twenty-five false to sixteen real. The fleet moves down by twenty-eight with nothing added, and eighteen of the twenty-eight are one copied preferences controller template repeated across nine apps.

A first draft of the purity walk was fail-open and silently cleared an unguarded case-dossier download, because the helper reached its store through a collaborator method whose name matches none of the gate's verb vocabularies. That was found by hand-reading the cleared set, not by any test I had written, and the walk now fails closed on any outbound call it cannot account for.

## Controls

Nine arms are pinned, and each is proven able to fail by breaking the specific half it guards: removing the helper exemption reddens the sanitiser arm, removing taint reddens the laundered-lookup arm, making the purity walk fail-open reddens the collaborator arm, and deeming every helper pure reddens two. The real-commit pair on opencatalogi's themes controller still discriminates exactly as it did before this change. The verb-object abuse controls from the earlier false-positive fixes all still pass, and the gate-57 arms go red when the old basename identity is restored.

Test counts move from 1142 to 1154 across the helper suites; the single pre-existing import error in the visual-coverage test file is present on main too and is untouched here.

## Deliberately not included

Two named weaknesses were measured and decided rather than inherited.

Matching guard patterns against comment-bearing text is real but has zero current exposure across all fifteen apps, so it is left for a standalone hardening change with its own arms.

Reading a forbidden status inside a catch block as an authorisation decision was rejected as framed. Seventy-four methods contain that shape but only fourteen would move, and twelve of those fourteen catch a typed authorisation exception and translate it, which is a correct delegated guard. Implementing it would have manufactured about a dozen false positives in exactly the failure mode this gate is trying to escape. The correct rule keys on the caught exception type and has a fleet exposure of one method.

Full write-up, including what could not be verified, is in the fleet-board findings file for this issue.

### DIFF
--- a/hydra-gates/scripts/lib/check_no_admin_idor.py
+++ b/hydra-gates/scripts/lib/check_no_admin_idor.py
@@ -1044,8 +1044,215 @@ _REQUEST_BOUND_ASSIGN_RE = re.compile(
     r"|->\s*getUploadedFile\s*\(|\$_(?:GET|POST|REQUEST|COOKIE|FILES)\b)"
 )
 
+# ---------------------------------------------------------------------------
+# Pattern 6, clause 2: which calls can actually RESOLVE AN OBJECT
+# (`ConductionNL/.github#398`)
+# ---------------------------------------------------------------------------
+#
+# Clause 2 vetoes when "a caller-controlled value reaches an unscoped call".
+# It used to treat EVERY method call as such a call, so a method that was
+# already correctly scoped was flipped to a finding by a line that touches no
+# store at all. Isolated in both directions on one body:
+#
+#     $safeKey = $this->sanitizeKey($key);       <- ADDING this line: 0 -> 1
+#     return $this->store->find($key, $uid);     <- unchanged, still scoped
+#
+# and adding a session identity to the SANITISER's own argument list cleared it
+# again — which is the tell that the veto, not the data access, was deciding.
+#
+# ⚠️ ADDING AN AUTHORISATION GUARD IS THE WRONG REPAIR for that shape. The
+# method is already scoped; a guard would be dead code written to satisfy a
+# misreading, and gate-7's whole credibility problem is that the fleet learned
+# to trust its silences.
+#
+# THE FIX IS IN TWO HALVES AND *BOTH* ARE REQUIRED. The second one alone is a
+# recall regression, which is the trap this class is laid over:
+#
+#   (A) TAINT PROPAGATION. A local assigned from an expression mentioning a
+#       caller-supplied value IS caller-supplied. Before this, `$safeKey` was
+#       invisible to clause 2 — so
+#
+#           $safeKey = $this->sanitizeKey($key);
+#           return $this->store->find($safeKey);      // REAL IDOR
+#
+#       was reported ONLY BY ACCIDENT, via the same veto that produced the
+#       false positive above. Exempting the sanitiser without (A) would have
+#       silently cleared that method. (A) is a recall INCREASE and lands first.
+#
+#   (B) A same-class `$this->helper(...)` that demonstrably resolves nothing
+#       does not, by itself, veto. The helper is looked up in this class's own
+#       source and ITS BODY IS READ — nothing is inferred from its name, for
+#       the same reason Pattern 4 refuses to infer anything from a property
+#       name. A body that performs any data access, any mutation or any
+#       request read is NOT clearable, and an unresolvable callee is NOT
+#       clearable. Fail-closed in every direction.
+#
+# With both, the sanitiser case clears and the IDOR case above still reports —
+# proven by the `SanitiserHelperFalsePositiveTest` near-miss arms.
+_TAINT_ASSIGN_RE = re.compile(
+    r"\$([A-Za-z_][A-Za-z0-9_]*)\s*=\s*([^;\n]{1,400});")
+_TAINT_FOREACH_RE = re.compile(
+    r"\bforeach\s*\(\s*([^)]{1,200}?)\s+as\s+(?:\$[A-Za-z_][A-Za-z0-9_]*\s*=>\s*)?"
+    r"\$([A-Za-z_][A-Za-z0-9_]*)")
 
-def _has_session_identity_handoff(body: str, params) -> bool:
+# PSR-3 logging is the other call shape that cannot resolve an object. The
+# interface is fixed, every method returns void, and a controller that puts a
+# caller-supplied id into its error message is doing diagnostics, not a
+# lookup. Measured while hand-reading the sample: openbuild's
+# `AppOverrideController::getUser` is correctly scoped
+# (`getUserDelta(appId: $appId, uid: $user->getUID())`) and would STILL have
+# been reported after (B), purely because of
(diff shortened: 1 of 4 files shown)


rubenvdlinde:
## Handover to whoever owns `run-hydra-gates.sh` / `bin/hydra-gates`

I did not touch either file — this is a consequence of the gate-57 change that belongs on the runner side.

With the foundation fail-safe now actually engaging in CI, the checker emits **no findings** for openregister and openconnector and explains why on **stderr**: the consumers of a foundation repo are not on disk, so deadness is unprovable. The runner sends stdout to the findings log and stderr to a separate `.err` file, then counts lines in the findings log. Zero lines means it reports gate-57 as passing.

So for those two repos gate-57 will now print a pass over a scope it explicitly declined to judge, and that pass counts toward the applicable-gate coverage line. That is the same shape as the gate-16 problem: a gate that inspected nothing prints the same word as a gate that looked and found nothing. Before this change the same two repos were printing eight and three findings, which was worse — the gate was answering a question it could not answer — but the new state is a false green rather than a false red, and false greens are the ones this fleet has learned not to notice.

The checker side is already set up to make this a one-line runner change: the skip line is stable and greppable, beginning with `[orphaned-write-capability] SKIP:` on stderr, and it names the app id and the reason. Suggested handling is to treat the presence of that marker in the `.err` file as a not-applicable verdict carrying the checker's own reason, rather than a pass — so the coverage block names it instead of silently counting it.

I have deliberately not implemented this, both because the file is yours and because the not-applicable versus pass distinction interacts with the coverage-enforcement flag in ways you are better placed to judge.

Separately and only if you are already in there: I could not run `test_gate_orphaned_capability_fixtures.sh`, the one suite that exercises gate-57 through the runner. The box was carrying about twenty concurrent gate processes from other agents at the time and the brief forbids adding another. Everything else in both suites was run and is green.

PR #420, opened 2026-08-13
fix(gate-50): a comment must neither consume the window nor satisfy it (#415)

### LINKED ISSUE #415: gate-50's 10-line window counts comment lines — 8 of 16 findings in opencatalogi were guards pushed out of range by the comment explaining them
## Summary

gate-50 looks for a fail-mode guard in a **10-line window** anchored at the end of the config read. In opencatalogi, **8 of its 16 findings were reads whose guard genuinely exists, 12 to 89 lines downstream** — and in the sharpest case the thing separating the read from its guard was *the comment explaining the guard*.

I fixed opencatalogi by moving each read to sit directly above its own check (ConductionNL/opencatalogi#867), which is better code and which I would do again. But the window will keep producing this class, and it will produce it **most often in the code that documents itself best** — which is the code this fleet asks people to write.

## The measurement

opencatalogi at `74e12c82`, canonical package `5e73e640`, full tree. For each of the 16 findings I measured the distance from the read to the nearest line matching gate-50's **own** guard vocabulary:

```
8 reads   guard 12-89 lines downstream    -> real guard, outside the window
8 reads   no guard within 60 lines        -> genuinely unguarded (all 8 fixed as real defects)
```

The eight false positives, with the distance and the guard that was already there:

| site | distance | guard |
|---|---:|---|
| `SetupController.php:299` / `:300` | 13 / 12 | `if ($publicationRegister !== '')`, `if ($publicationSchema !== '')` |
| `CatalogiService.php:371` / `:372` | 22 / 21 | `if ($schema === '' \|\| $register === '')` + `logger->error` |
| `DirectoryService.php:2258` / `:2259` | 27 / 26 | `if (empty($listingSchema) === false && empty($listingRegister) === false)` |
| `DirectoryService.php:2260` / `:2261` | 89 / 88 | `if (empty($catalogSchema) === false && empty($catalogRegister) === false)` |

🔑 **`CatalogiService::getCatalogFilters` is the one that matters.** Its guard is 22 lines away because a **20-line comment explaining why the guard is necessary** sits between the read and the check. That comment is exactly what this fleet's review culture asks for, and it is what made the gate call the site unguarded.

Same repo, same file: `DirectoryService.php:277-284` is the identical pattern with the guard **6** lines away, and it passes. So the gate is not measuring whether a guard exists; **it is measuring how much prose the author wrote between the two.**

## Why this matters more than a normal false-positive rate

The gate's own source already carries the argument against unclosable findings (#252, quoted in the `_ARRAY_VALUE` exemption): *"A finding here names a line at which no closing action exists… 47 of them in one repo is how a gate teaches people to stop reading it."* A finding whose only remedy is *move your comment* is close to that shape — the code is already correct, and the gate is asking for a formatting change.

And the failure direction is the dangerous one for adoption: eight of sixteen is a **50% false-positive rate on first contact**, which is how a gate's future silences stop being believed.

## Suggestions, in order of preference

1. **Skip comment and blank lines when counting the window.** Ten lines of *code* is the intent; ten lines of *text* is what is implemented. This is a small change to the window construction and it fixes every case above except the 88/89 pair, without loosening the rule for the single-line shape the gate was built on.
2. **Widen the window to the end of the enclosing statement block or ~30 lines**, whichever is shorter.
3. Failing either, **report the distance in the finding** (`"guard found 22 lines away, outside the 10-line window"` vs `"no guard found"`), so a reader can tell a formatting finding from a security finding without opening the file. Right now the two are byte-identical in the log, and that is what cost the triage time here.

## Sibling

`.github#358` records gate-26 being fooled **by** comments (a comment saying you have not written a baseline satisfied the check for having written one). This is the same seam from the other side: one gate reads comments it should ignore, the other is defeated by their length. Worth a single pass over both windows with `js_comment_mask`-style handling on the PHP side.

rubenvdlinde:
Closes #415.

gate-50's guard window was raw file text, so comment lines **both consumed the budget and supplied the evidence**. One cause, two failures pointing opposite ways.

| arm | before | after |
|---|---|---|
| textbook guard, 12 comment lines above it | **FAIL** (false positive) | PASS |
| unguarded `api_token`, TODO naming the missing guard | **PASS** (false negative) | FAIL |
| clean guard, no comments | PASS | PASS |
| genuinely unguarded read | FAIL | FAIL |

**The false negative is the serious half.** An unguarded read reported clean because the TODO above it read *"we should throw new RuntimeException here when empty, and compare `$tok === ''` before use. Not done yet."* — a comment **stating the debt** satisfied the gate that exists to collect it. That is the defect gate 19 was fixed for; nobody looked for it here.

**The false positive is the one that erodes the gate.** 8 of gate-50's 16 opencatalogi findings were real guards pushed out of range by the comment explaining them. The gate penalised the documented fix and passed the undocumented one — an author who deletes the comment goes green having changed nothing about the code's safety.

## Fix

Strip comment **text** before building the window: comments occupy no budget and offer no evidence. Budget becomes eleven lines of **code**, not eleven lines of file. Line numbers are preserved (text blanked, lines kept) because findings report one.

String-safe rather than a naive split — `'https://x'` must not truncate a line, `#[PublicPage]` is a PHP 8 attribute. Quote state resets per line, so an unterminated literal corrupts at most its own line.

## Proof

New arms C14/C15/C16. **With the runner change reverted, C14 and C15 both flip and C16 does not** — so the two new arms detect this regression, and the third is the control against a stripper that eats code rather than comments. Suite 42 → 45 assertions, all green.

Half of every gate-50 count fleet-wide was suspect until this landed.

### DIFF
--- a/hydra-gates/scripts/lib/test_gate_45_to_55_acceptance.sh
+++ b/hydra-gates/scripts/lib/test_gate_45_to_55_acceptance.sh
@@ -539,6 +539,83 @@ _outC13="${_tmp}/c13.txt"
 _run "${_appC}" "${_outC13}"
 _expect "${_outC13}" 50 "FAIL" "does not mistake an arrow closure's => for an array key"
 
+# C14 / C15 — A COMMENT MUST NEITHER CONSUME THE WINDOW NOR SATISFY IT (#415).
+#
+# One cause, two failures pointing opposite ways, so both arms are needed: a
+# fix for either one alone can be had by moving the window boundary, and that
+# makes the other worse.
+#
+# C14 (was a FALSE POSITIVE): a textbook guard, reported unsafe because twelve
+# lines of ordinary explanation sat between it and the call. 8 of gate-50's 16
+# opencatalogi findings were this shape. The gate penalised the DOCUMENTED fix
+# and passed the undocumented one — an author who deletes the comment goes
+# green having changed nothing about the code's safety.
+_write_service '    public function listing(): array
+    {
+        $reg = $this->config->getValueString(Application::APP_ID, '"'"'listing_register'"'"', '"'"''"'"');
+        // We deliberately do not fail hard on a missing register: it is
+        // optional in single-tenant installs, the directory sync job re-runs
+        // nightly, and a missing value therefore self-heals within a day.
+        // See ADR-012 for the full rationale, and the migration plan that
+        // removes this branch entirely once the 3.x provisioning path has
+        // landed in every deployment we currently support. Until then the
+        // empty case must return an empty listing rather than raise, because
+        // the public directory endpoint is unauthenticated and a raise there
+        // is an information leak of its own. Twelve lines is not an unusual
+        // amount of explanation for a decision with that many moving parts,
+        // and none of it changes what the code below does.
+        //
+        // The actual guard follows.
+        if ($reg === '"'"''"'"') {
+            return [];
+        }
+
+        return $this->load($reg);
+    }'
+_commit "${_appC}" "a guarded read with a long comment between read and guard"
+_outC14="${_tmp}/c14.txt"
+_run "${_appC}" "${_outC14}"
+_expect "${_outC14}" 50 "PASS" "a comment between the read and its guard does not consume the window"
+
+# C15 (was a FALSE NEGATIVE, and the more serious half): an unguarded read of
+# an api_token reported CLEAN because the TODO above it happened to contain
+# the words the guard regex looks for. A comment STATING THE DEBT satisfied
+# the gate that exists to collect it — the same defect gate 19 was fixed for,
+# never looked for here. Note the fixture's comment contains BOTH a `throw
+# new` and a `=== ''`, so it exercises two of the six guard arms at once.
+_write_service '    public function push(): void
+    {
+        $tok = $this->config->getValueString(Application::APP_ID, '"'"'api_token'"'"', '"'"''"'"');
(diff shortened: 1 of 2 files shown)


rubenvdlinde:
Correcting the impact claim in this PR body after measuring it.

I wrote "half of every gate-50 count fleet-wide is suspect until this landed." Measured, old checker vs new, over `lib/**/*{Controller,Service}*.php`:

| repo (current `development`) | before | after | Δ |
|---|---:|---:|---:|
| docudesk · procest · softwarecatalog · larpingapp | 0 | 0 | 0 |
| opencatalogi | 4 | 4 | 0 |
| **opencatalogi @ `74e12c82`** (the tree #415 was filed on) | **16** | **12** | **−4** |

Two corrections:

- **#415's "8 of 16" is 4, not 8** — `SetupController:299,300` and `CatalogiService:371,372`, all four genuine guards pushed out of the window by the comment explaining them.
- **The false-negative half has zero live instances across five repos.** It is real and fixture-proven, but it is currently costing nothing. The value of this fix is prospective — it stops the class recurring — not a live cleanup, and I implied otherwise.

The rig was positive-controlled before I believed the zeros (five identical verdicts is the shape of a broken instrument): the same two checkers over the acceptance fixtures give B 0→1 and E 1→0. Layer: this is the checker over a file list I selected, not the gate through CI; it is corroborated by opencatalogi's 16 matching the figure CI reported on that tree.

The fix and its arms stand. Only the impact sentence was wrong.

Issue #196, opened 2026-08-06
gate-5 (route-auth) is satisfied by a COMMENT mentioning the attribute — a false-negative vector in a security gate


rubenvdlinde:
## What

gate-5 decides whether a routed method has an auth attribute by grepping a window of source lines:

```sh
head_block=$(sed -n "${start},${def_line}p" "$path")
if ! echo "$head_block" | grep -qE '#\[(PublicPage|NoAdminRequired|NoCSRFRequired|AuthorizedAdminSetting)\b|@(PublicPage|NoAdminRequired|NoCSRFRequired)\b'; then
```

That window is raw text. It does not distinguish an **attribute** from the same characters appearing inside a **docblock comment**.

So this passes the gate:

```php
/**
 * ADMIN ONLY: `#[NoAdminRequired]` is deliberately NOT used here.
 */
public function analytics(string $productId): JSONResponse
```

— while the method has no attribute at all.

## Why it matters

This is the **expensive direction** for a security gate: a pass leaves no log, so a method that mentions any of the four attribute names in prose is silently exempt. Anyone can turn the gate off for a method by writing about it. The gate's own header already records finding a false negative of a different shape (the 20-line lookback borrowing a neighbour's attribute) and calls that out as "the expensive direction, and invisible because a pass leaves no log" — this is the same class.

## Where it shows

openconnector's `ProductSubscriptionsController`:

- `subscribe()` — no attribute; passes the gate solely because its docblock says `` `#[NoAdminRequired]` is deliberately NOT used here ``
- `analytics()` — no attribute, no such sentence; **flagged** (ConductionNL/openconnector#1165)

Two methods with identical auth posture, one flagged and one not, decided entirely by prose.

## The awkward part

There is no good way for an author to *positively* declare "admin only" on a plain controller method: absence of `#[NoAdminRequired]` **is** the admin gate, and `@NoAdminRequired` in a docblock would make it non-admin. So documenting the posture is genuinely the only option available — and it happens to also satisfy the regex, which is why this has not been noticed.

## Suggested direction

Strip comments before matching, and give the "deliberately admin-only" case an explicit, greppable marker that is not a real attribute name — e.g. an `@auth admin-only <reason>` tag in the same family as `@spec exclude` / `@e2e exclude`. Then:

- an attribute is an attribute
- a deliberate admin-only endpoint is declared, not inferred from a missing line
- prose stops being load-bearing

[...]

PR #427, opened 2026-08-13
fix(gate-7): a (string) cast blinded Pattern 6 — the guard was unchanged, only its spelling (#414)

### LINKED ISSUE #414: gate-7 Pattern 6 is blinded by a (string) cast on the session identity — a false positive whose recommended repair adds a redundant guard
## Summary

gate-7's session-identity recogniser (Pattern 6) is **blinded by a `(string)` cast** on the identity expression. The guard is unchanged; only its spelling moves the verdict.

Found on openbuild `AppOverrideController::getUser()`, an endpoint whose UID is never a request parameter and is handed straight to `AppOverrideService::getUserDelta(appId:, uid:)` — textbook Pattern 6, flagged anyway.

## Four arms on the shipped file, nothing else changed

| arm | expression carrying the session identity | `check_no_admin_idor.py` |
|---|---|---|
| **A** (shipped) | `uid: (string)$user->getUID()` | **FAIL** |
| B | `uid: $user->getUID()` | PASS |
| C | `$uid = (string)$user->getUID();` … `uid: $uid` | **FAIL** |
| D | `appId: (string)$appId, uid: $user->getUID()` | PASS |
| E | `$uid = $user->getUID();` … `uid: $uid` | PASS |

Both directions demonstrated: the checker reports on this file (arm A), and stops reporting on a one-token change (arm B).

- **D rules out** "any cast in the call defeats it" — casting an *unrelated* argument is harmless.
- **C rules out** "hoist it into a local" as the workaround — the cast defeats the recogniser *through a variable too*. C is the arm I expected to pass, and it did not; the expectation was computed before the run.

## Why this class matters more than its count

It is a **false positive that recommends the wrong repair**. The runner's own FAIL guidance says *"The fix is to scope the object to the caller"* — but the object **is** already scoped to the caller. Following the guidance means adding a redundant guard to an endpoint that already had one, and then believing the gate about it.

`(string)$user->getUID()` is a common defensive spelling precisely because `IUser::getUID()` carries no PHP return type (only `@return string`, `lib/public/IUser.php`), so static analysers encourage the cast. **The gate and the analysers pull in opposite directions on the same line.**

## Measured blast radius — reported as measured, not extrapolated

Across the six apps in this sweep (`openbuild`, `doriath`, `scholiq`, `launchpad`, `larpingapp`, `openconnector`):

- the `(string)`-cast session-identity spelling occurs **7 times** in `lib/` (openbuild 5, launchpad 2, the other four 0);
- but only **1 of the 20** live gate-7 findings is this class.

⚠️ The other 12 fleet apps are **unmeasured**. The remaining 6 sites in these six apps sit in methods that pass for other reasons, so 7 is an upper bound on the sites, not on the findings.

## Fixed in-app meanwhile

ConductionNL/openbuild#196 drops the cast (it is a no-op: `getUserDelta()` declares `string $uid` under `strict_types=1`). That is the right change for that repo either way, but it is a workaround for this defect, not a fix for it — the next app to write the idiomatic cast gets the same false positive.

## Suggested fix

Normalise the identity expression before matching: strip a leading cast (`(string)`, `(int)`) when tracing an argument back to a `getUID()` / `getUser()->getUID()` source, and apply the same normalisation on the local-variable trace so arms C and E agree.

rubenvdlinde:
Closes #414.

## Reproduced first — five arms on the shipped openbuild file, one token varying

| arm | expression carrying the session identity | gate-7 |
|---|---|---|
| **A** (shipped) | `uid: (string)$user->getUID()` | **FAIL** |
| B | `uid: $user->getUID()` | PASS |
| **C** | `$uid = (string)$user->getUID();` … `uid: $uid` | **FAIL** |
| D | `appId: (string)$appId, uid: $user->getUID()` | PASS |
| E | `$uid = $user->getUID();` … `uid: $uid` | PASS |

D rules out "any cast in the call defeats it" — an unrelated cast is harmless. C rules out "hoist it into a local" as the workaround.

## Mechanism

Pattern 6's identity recogniser applies control 2 — *"the expression may carry no call ARGUMENTS"* — with `\(\s*[^)\s]`. A PHP cast is written with exactly those bytes:

```python
>>> re.search(r"\(\s*[^)\s]", "(string)$user->getUID()")   # matches "(s"
```

So the guard did not change; only its spelling did. And the cast is not an odd spelling: `IUser::getUID()` carries no PHP return type (only `@return string`), so Psalm and PHPStan actively encourage writing it. **The gate and the analysers pulled in opposite directions on the same line.**

This is a false positive whose recommended repair is **wrong**: gate-7's FAIL text says *"the fix is to scope the object to the caller"* and the object already was — the UID is never a request parameter. Following the guidance means adding a redundant guard to an endpoint that had one, and then believing the gate about it.

## The fix

A leading `(string)` / `(int)` / `(integer)` cast is normalised away before classification. `(array)` and `(object)` are deliberately absent — they would launder a payload into an identity. Both surviving controls run **after** the strip, so nothing widens: `(string)canAccess($id, $uid)` still has a real argument list, `(string)$row['ownerId']` still has a subscript.

⚠️ The strip is applied in `_argument_is_session_identity` **as well**, before the bare-variable test — not only inside `_is_identity_expression`. Doing it in one place only would have routed `(string)$targetUid` past the declared-parameter veto (`_IDENTITY_TOKEN_RE` matches "uid"), turning a false positive into a false **negative**. That is the direction that leaves no log to notice.

## Measured before/after — gate-7 on `lib/Controller`, all 18 core apps

| repo | before | after | Δ |
|---|---|---|---|
| decidesk | 15 | 15 | 0 |
| docudesk | 23 | 23 | 0 |
| doriath | 4 | 4 | 0 |
| hermiq | 17 | 17 | 0 |
| larpingapp | 1 | 1 | 0 |
| launchpad | 1 | 1 | 0 |
| nldesign | 0 | 0 | 0 |
| openbuild | 6 | 6 | 0 |
| opencatalogi | 14 | 14 | 0 |
| openconnector | 80 | 80 | 0 |
| openregister | 95 | 95 | 0 |
| pipelinq | 50 | 50 | 0 |
| portaliq | 1 | 1 | 0 |
| procest | 27 | 27 | 0 |
| scholiq | 4 | 4 | 0 |
| shillinq | 18 | 18 | 0 |
| softwarecatalog | 30 | 30 | 0 |
| zaakafhandelapp | 89 | 89 | 0 |
| **TOTAL** | **475** | **475** | **+0** |

Zero delta is the honest number, and it is not zero effect: ConductionNL/openbuild#196 had already dropped the cast in-app, so the only live site was gone before this ran.

**Two-directional control on the byte-exact pre-#196 file**: WITHOUT this change it reports `method=getUser rule=no-auth-guard-in-body` (**7** findings); WITH it, it does not (**6**). That matches openbuild#196's own recorded measurement of 7 → 6 exactly. Same file, same checker, one token.

## Revert control — which arms are evidence

Suite re-run with the **fixtures kept and the checker reverted**:

- **EVIDENCE** — `authn-vs-authz/clean` `handoffCastIdentity()` + `handoffCastLocal()`: the clean arm FAILS with `2 method(s) with NoAdminRequired + no guard` → **130 passed / 4 failed**. With the fix: **134 / 0**. (A false-positive fix shows up as a clean arm that stops failing.)
- **CONTROL** — `authn-vs-authz/planted` `castCallerValue()`: a cast on a value the *caller* chose. Reports before **and** after, so "strip every cast and classify what is left" cannot pass this bundle.
- **CONTROL** — the three `#365` planted arms and the four original clean shapes: unchanged in both directions.

`test_check_no_admin_idor.py`: **145 pass**, seven of them new (including `test_cast_on_a_caller_supplied_value_is_still_reported`, `test_cast_does_not_make_object_data_an_identity`, `test_array_and_object_casts_are_not_stripped`). `test_gate_orphaned_capability_fixtures.sh` fails identically on pristine `main` (missing fixture dirs) — pre-existing, not this branch.

### DIFF
--- a/hydra-gates/scripts/lib/check_no_admin_idor.py
+++ b/hydra-gates/scripts/lib/check_no_admin_idor.py
@@ -728,6 +728,44 @@ def _top_level_split(text: str, sep: str) -> list:
     return [p.strip() for p in parts]
 
 
+# A LEADING SCALAR CAST, stripped before the expression is classified
+# (`ConductionNL/.github#414`).
+#
+# 🔴 THE SPELLING MOVED THE VERDICT, AND THE GUARD DID NOT. Control 2 below
+# rejects "any call with arguments" with `\(\s*[^)\s]` — and a cast is written
+# with exactly those bytes, so `(string)$user->getUID()` read as a call taking
+# an argument and stopped being an identity. Measured on openbuild's
+# `AppOverrideController::getUser()`, a method whose uid is never a request
+# parameter, on one file with one token varying:
+#
+#     uid: (string)$user->getUID()          FAIL
+#     uid: $user->getUID()                  PASS
+#     $uid = (string)$user->getUID(); …     FAIL   <- through a local, too
+#     appId: (string)$appId, uid: $user…    PASS   <- an unrelated cast is fine
+#
+# That is a FALSE POSITIVE WHOSE RECOMMENDED REPAIR IS WRONG: gate-7's FAIL
+# text says "scope the object to the caller", and the object already was. And
+# the cast is not a quirk — `IUser::getUID()` carries no PHP return type, only
+# `@return string`, so Psalm/PHPStan actively encourage writing it.
+#
+# ⚠️ WHAT IS *NOT* STRIPPED, and why the list is short. `(array)` and
+# `(object)` would launder a payload into an identity, so they are absent. The
+# two surviving controls do the rest of the work AFTER the strip, which is what
+# keeps this from widening anything: `(string)canAccess($id, $uid)` still has a
+# real argument list and `(string)$account['ownerId']` still has a subscript,
+# so both stay non-identities.
+_LEADING_SCALAR_CAST_RE = re.compile(r"^\(\s*(?:string|int|integer)\s*\)\s*")
+
+
+def _strip_leading_scalar_casts(expr: str) -> str:
+    """Remove leading ``(string)`` / ``(int)`` casts — see `#414` above."""
+    while True:
+        stripped = _LEADING_SCALAR_CAST_RE.sub("", expr, count=1)
+        if stripped == expr:
+            return expr
+        expr = stripped
+
+
 def _is_identity_expression(expr: str, context: str = "") -> bool:
     """True when *expr* reads as "the caller's identity" and nothing else.
 
@@ -735,8 +773,11 @@ def _is_identity_expression(expr: str, context: str = "") -> bool:
     ARGUMENTS and no array subscript, so a guard call (``canAccess($id, $uid)``)
     and object data (``$account['ownerId']``) can never be read as an identity
     however their names are spelled.
+
+    A leading scalar cast is normalised away first (`#414`) — it is a SPELLING
+    of the same expression, and control 2 could not tell it from a call.
     """
-    expr = expr.strip()
+    expr = _strip_leading_scalar_casts(expr.strip())
     if expr == "":
         return False
     if expr[0] in ("'", '"'):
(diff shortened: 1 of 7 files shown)


rubenvdlinde:
Merged `main` in and resolved.

**#426 and #427 each merged cleanly onto `main` individually and conflicted with each other** — both appended a test class at the same point in `test_check_no_admin_idor.py`. Testing each against the base could never have surfaced that; it only appears when you build the combined tree.

The conflict is purely additive, so the resolution is union, not a choice. Verified on the combined tree before pushing:

- `test_check_no_admin_idor` — **162 tests, OK** (the #425 + #426 + #427 combination, which no run had exercised)
- `test_gate_acceptance_matrix.sh` — **137 passed, 0 failed, 60 of 65 gates fixtured**

Both checkers compile. Re-running CI on the merged head now.

Issue #340, opened 2026-08-11
gate-53 removals-invariant equates reachability with the static .menu tree — 8 procest removals whose replacement is a folderSidebar/viewMode/sidebar-tab are reported as functionality loss


rubenvdlinde:
## Summary

`gate-53` (effective-manifest-crossref), check `removals-invariant`, reports 8 errors on **ConductionNL/procest** that describe a menu topology, not a functionality loss. ADR-044's invariant is *no functionality loss*; the gate's proxy for it is *"some surviving entry in the static `.menu` tree reaches this route"*. Those two come apart the moment a page is reached dynamically, and procest has two shipped, deliberate architecture changes that do exactly that.

## The findings

```
/menu-layout/removals/0: removal 'BezwaarBeroepGroup' orphans route 'BezwaarBeroepOverview'
/menu-layout/removals/1: removal 'Bezwaren'            orphans route 'Bezwaren'
/menu-layout/removals/2: removal 'Beroepen'            orphans route 'Beroepen'
/menu-layout/removals/3: removal 'SubsidiesGroup'      orphans route 'Subsidies'
/menu-layout/removals/4: removal 'CaseMap'             orphans route 'CaseMap'
/menu-layout/removals/5: removal 'Voorstellen'         orphans route 'Voorstellen'
/menu-layout/removals/6: removal 'Advice'              orphans route 'Advice'
/menu-layout/removals/7: removal 'BesluitvormingAgenda' orphans route 'AgendaCompiler'
```

## Why they are not losses

Two sequential, documented changes replaced static top-level nav with dynamic reachability:

1. **`case-type-navigation`** (commit `d6824aa20`) — the standalone Bezwaar / Beroep / Subsidie nav was replaced by case-type children resolved at runtime onto the `Cases` page's own `folderSidebar`, and `Cases` gained a `map` viewMode that supersedes the standalone `CaseMap` leaf.
2. **`consume-decidesk-besluitvorming-leaf`** (commit `d2df51b8b`) — the standalone Besluitvorming nav (Voorstellen / Advice / Agenda) was replaced by a `BesluitvormingLeafTab` sidebar tab on `CaseDetail`.

Both replacements are live and menu-reachable via `Cases` / `CaseDetail`. The gate walks only the static `.menu` tree, so a runtime `folderSidebar` filter, a page `viewMode`, and a per-object sidebar tab are all invisible to it as reachability.

[...]

PR #434, opened 2026-08-13
fix(gates): gate-3 spares a contract-imposed unused param; gate-53 compares page identity not route spelling

### LINKED ISSUE #339: gate-3 caller-identity-ignored: an interface-contract $userId an implementor correctly ignores has no app-side fix
## Summary

`gate-3` (stub-scan), rule `caller-identity-ignored`, reports three findings on **ConductionNL/procest** for which **no correct app-side fix exists**. The parameter it objects to is an interface-contract obligation, and the interface's other implementors genuinely use it.

This is the same family as the interface-declaration false positive already fixed in #291 — but one step further along: #291 covered *bodiless declarations*, and this is an *implementation with a real body that correctly does not need the parameter*.

## The findings

```
lib/Service/Transitions/ChecklistGuard.php:65       method=evaluate rule=caller-identity-ignored param=$userId
lib/Service/Transitions/RequiredDocumentGuard.php:50 method=evaluate rule=caller-identity-ignored param=$userId
lib/Service/Transitions/RequiredFieldGuard.php:48   method=evaluate rule=caller-identity-ignored param=$userId
```

## Why there is no fix the author can make

All three implement one interface:

```php
// lib/Service/Transitions/GuardEvaluatorInterface.php:47
public function evaluate(array $guardConfig, array $case, string $userId): GuardResult;
```

There are five implementors, and `$userId` references inside `evaluate()`:

| implementor | `$userId` refs | uses caller identity? |
|---|---|---|
| `RoleGuard` | 12 | yes — this is the whole point of the class |
| `MandaatGuard` | 3 | yes |
| `ChecklistGuard` | 1 (signature only) | no — inspects the case's checklist |
| `RequiredDocumentGuard` | 1 (signature only) | no — inspects the case's documents |
| `RequiredFieldGuard` | 1 (signature only) | no — inspects a case field |

A guard that asks "is this required field filled in?" has no business consulting the caller. The three flagged classes are *correct*. The only ways to silence the gate are:

1. **delete the parameter** — breaks `GuardEvaluatorInterface` and every call site in `GuardRegistry::evaluateAll()`; or
2. **reference `$userId` pointlessly** in the body to get the grep count to 2 — i.e. write dead code to satisfy a stub detector.

Both make the code worse. Neither is something a reviewer should accept.

## The signal the gate is walking past

All three already carry an explicit, reason-bearing marker that the parameter is deliberately unused:

```php
 * @param string $userId Current user UID (unused)
 *
 * @SuppressWarnings(PHPMD.UnusedFormalParameter)
```

That is the house convention for exactly this situation, it is already enforced elsewhere (PHPMD reads it), and gate-3 does not look at it.

## Suggested fix

In the `caller-identity-ignored` loop in `run-hydra-gates.sh`, skip a method when **either** holds:

- the docblock immediately preceding the signature carries `@SuppressWarnings(PHPMD.UnusedFormalParameter)`, **or**
- the `@param` line for the flagged parameter carries an explicit unused/ignored marker.

The first is preferable: it is an existing, tool-enforced annotation rather than a new gate-specific opt-out, so it cannot be sprinkled around freely without PHPMD agreeing.

Please guard the change with a planted true positive: a service method that takes `$userId`, ignores it, and carries **no** suppression must still be reported. An exemption that cannot fail is worth less than the finding it removes.

## Impact

3 findings on procest. Fleet-wide this pattern will appear anywhere a strategy/guard/handler interface passes caller identity to implementors that do not all need it, which is a normal shape.

rubenvdlinde:
Closes #339. Partially addresses #340 — see "What #340 actually is" below; that issue needs a decision, not just this patch.

Both issues were filed with the observation that **no app-side fix exists**. Rule 1 was to test that claim before writing anything, so both were reproduced first, on a fresh `development` checkout of six fleet apps.

## Reproduction — both still live

Full 60-gate run, `hydra-gates@c26f9a3`, apps at `development`:

| app | gate-3 `caller-identity-ignored` | gate-53 `removals-invariant` |
|---|---|---|
| procest `792fe1f4b` | **3** | **8** |
| opencatalogi | 0 | 0 |
| openregister | 0 | 0 |
| softwarecatalog | 0 | 0 |
| docudesk | 0 | 0 |
| larpingapp | 0 | 0 |

Both reproduce exactly as filed, at the same three files and the same eight removals. Both live populations are **procest only**. `menu-layout.json#removals` is non-empty in **1 of the 18 core apps** (surveyed via the API on `development`; the probe's positive control is that it returned procest's 8).

## gate-3 — outcome: the question is right, the detection is incomplete. Fixed.

The rule exists for decidesk#45: a builder-generated `authorize*()` stub that accepts the caller and forgets to check it. That is a real defect and the rule should keep finding it.

It cannot currently tell that shape apart from a correct strategy implementation. All three procest findings implement `GuardEvaluatorInterface::evaluate(array, array, string $userId)`; two of the five implementors use the parameter, which is the whole point of those classes. A guard answering "is this required field filled in?" has no business consulting the caller. Deleting the parameter breaks the interface and every call site; referencing it pointlessly is dead code written to satisfy a stub detector. **Neither edit is a fix**, which is why procest reported instead of closing.

New helper `check_caller_identity_exempt.py`. A finding is exempt only when **both** hold:

1. **The signature is imposed from outside** — a supertype resolvable *in this repository* (`implements`/`extends`, transitively) declares the same method with the same parameter. Fully mechanical; nothing the author asserts. This is what makes the exemption un-sprinklable: an invented service method has no supertype, so a fixer agent cannot escape the rule by adding a docblock line.
2. **The author marked that parameter unused** — the docblock's `@param` line for the flagged parameter carries an explicit unused marker.

Condition 1 alone would exempt a genuinely gutted implementation (a `RoleGuard` that stopped consulting `$userId` is contract-imposed too, and that *is* the defect). Condition 2 alone is a one-word escape hatch.

#339 suggested keying on `@SuppressWarnings(PHPMD.UnusedFormalParameter)` and preferred it. I took the stricter of the two options it offered: that annotation is **method-wide and cannot name which parameter it excuses**. procest's `MandaatGuard` carries it — for a different parameter — while genuinely using `$userId`. Keying on it would have pre-authorised MandaatGuard's `$userId` for a reason that was never about `$userId`. The per-`@param` marker is per-parameter, and all three findings already carry it, so no app edit is needed.

Fail-closed throughout: a missing helper, an unreadable file, or a supertype outside this repository all leave the finding standing.

### Acceptance — `test_gate_3_contract_param.sh`, 5 arms, 4 of them controls

| arm | fixture | expected |
|---|---|---|
| 1 | contract + marker (procest's shape) | 0 findings |
| 2 **control** | contract, marker deleted | still reported |
| 3 **control** | marker, no contract (decidesk#45 with a docblock) | still reported |
| 4 **control** | both look present, interface file absent from the repo | still reported (fail-closed) |
| 5 **control** | supertype declares the method **without** the parameter | still reported |

Arms 2–5 each delete exactly one half of the exemption. **Revert test**: the same suite run against `origin/main`'s unmodified runner — arm 1 goes red (`RequiredFieldGuard.php:16 method=evaluate rule=caller-identity-ignored param=$userId`), all four controls stay green, i.e. the controls were never depending on the fix.

**Effect**: procest 3 → 0 across a full 60-gate run, and gate-3 is the **only** verdict that moved (runner exit 8 → 7). 0 → 0 on the other five apps.

### The detection defect (fixed)

`menu[].route` may hold **either** a `pages[].id` **or** a `pages[].route` — check (a) in the same file accepts both, and apps use both spellings. `removals-invariant` compared the raw strings, so two menu entries reaching the *same page* by different spellings did not count as reaching each other. Demonstrated on a two-entry manifest:

```
menu:  {id: items-by-id,   route: "ItemsPage"}
       {id: items-by-path, route: "/items"}
pages: {id: "ItemsPage", route: "/items"}
removals: ["items-by-path"]
```

→ `removal 'items-by-path' orphans route '/items' — no surviving menu entry reaches it`

That is precisely the "duplicate navigation entry whose page is still reachable" ADR-044 §5 sanctions, reported as a loss. Both sides now resolve to the page id before comparison. An unresolvable reference is left as-is, so two broken entries cannot vouch for each other (check (a) already fails those).

Two arms added to `test_check_manifest_crossref.js`, off one fixture pair differing only in whether the second entry survives:
- id-spelled survivor covers a path-spelled removal → 0 errors
- **control**: sole entry removed, nothing else reaches the page → still exactly 1 error

Existing `broken/` fixture arm (`cases-index`) unchanged and still red. Whole suite green (28 assertions).

**Fleet A/B**, crossref findings before → after: procest 9→9, opencatalogi 0→0, openregister 0→0, softwarecatalog 1→1, docudesk 0→0, larpingapp 1→1. **Nothing moved** — this is a latent defect, not a live one, because only procest uses `removals` at all and its eight are not this shape.

### What #340 actually is — flagged for a decision, not patched

procest's 8 are **not** closable by this fix and I did not implement the enhancement #340 proposes, for three measured reasons. Detail in the issue comment; summary:

1. **A correct app-side fix does exist**, and #340 missed it. All eight ids are declared in procest's *own* `manifest.json` / `manifest.d/` fragments. Deleting the eight **menu** nodes at source (keeping the **pages** entries, which is what makes the routes routable) and emptying `removals` produces a **byte-identical effective manifest** — measured — and takes gate-53 from 8 errors to 0. So the finding is closable, behaviour-neutrally, without restoring the anti-pattern or retiring any route.
2. **That reveals the gate's real defect**: `removals-invariant` polices the *key*, not the *invariant*. The same menu outcome reached by editing a fragment is completely unchecked — I just did it eight times and got a PASS. The gate penalises the honest, centrally-documented authoring path and ignores the silent one, which is the opposite of what ADR-044 says it is guarding ("the hard risk … is **silently** losing a page").
3. **The enhancement #340 proposes would close at most 3 of the 8.** The `folderSidebar` join is genuinely mechanical; `viewMode` is a name heuristic, not a join; and the sidebar-tab arm is **not resolvable in principle** — `BesluitvormingLeafTab` resolves *another app's* component from the runtime `window.OCA.OpenRegister.integrations` registry, and nothing static links it to `Voorstellen`/`Advice`/`AgendaCompiler`. Implementing it would add PASSes for 3 and leave 5 red, which is the failure mode #340 itself asks to avoid.

Also worth recording: ADR-044 §5 says `removals` may retire **only** a duplicate navigation entry whose page is still reachable. The gate implements that faithfully. procest's eight are *supersessions*, not duplicates — so what #340 is really asking for is an amendment to ADR-044, which is an architecture call, not a gate call.

## Files

- `hydra-gates/scripts/lib/check_caller_identity_exempt.py` (new)
- `hydra-gates/scripts/lib/test_gate_3_contract_param.sh` (new)
- `hydra-gates/scripts/run-hydra-gates.sh` (gate-3 call site, fail-closed)
- `hydra-gates/scripts/lib/check_manifest_crossref.js` (removals-invariant normalisation)
- `hydra-gates/scripts/lib/test_check_manifest_crossref.js` (2 arms)

`source_scope.py` untouched. No app repository is changed by this PR.

### DIFF
--- a/hydra-gates/scripts/lib/check_caller_identity_exempt.py
+++ b/hydra-gates/scripts/lib/check_caller_identity_exempt.py
@@ -1,0 +1,272 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: EUPL-1.2
+"""check_caller_identity_exempt.py — is a gate-3 caller-identity finding closable?
+
+WHY THIS EXISTS (.github#339)
+-----------------------------
+gate-3's ``caller-identity-ignored`` rule reports any public method under
+``lib/Service`` / ``lib/Controller`` that declares ``$uid`` / ``$callerUid`` /
+``$userId`` / ``$caller`` and never references it in its body.  That is
+decidesk#45's shape: a builder-generated ``authorize*()`` stub that accepts the
+caller and then forgets to check it.
+
+It is ALSO the shape of a correct strategy implementation.  Measured on procest
+(``development`` @ ``792fe1f4b``), all three findings the gate produced were of
+the second kind::
+
+    lib/Service/Transitions/ChecklistGuard.php:64        method=evaluate param=$userId
+    lib/Service/Transitions/RequiredDocumentGuard.php:49 method=evaluate param=$userId
+    lib/Service/Transitions/RequiredFieldGuard.php:47    method=evaluate param=$userId
+
+All three implement ``GuardEvaluatorInterface::evaluate(array, array, string
+$userId)``.  Two of the five implementors (``RoleGuard``, ``MandaatGuard``) use
+the parameter — it is the whole point of those classes.  A guard that asks "is
+this required field filled in?" has no business consulting the caller, so the
+only edits that silence the gate are deleting the parameter (breaking the
+interface and every call site in ``GuardRegistry::evaluateAll()``) or
+referencing it pointlessly to get the grep count to 2.  Both make the code
+worse; the finding had no closing action.
+
+WHAT MAKES A FINDING EXEMPT
+---------------------------
+Two conditions, and BOTH must hold.  Either alone is insufficient, and that is
+deliberate:
+
+1. **The signature is imposed from outside.**  The declaring class names a
+   supertype (``implements`` / ``extends``, transitively) that is resolvable in
+   this repository and declares the same method with the same parameter.  This
+   is fully mechanical: nothing the author asserts, only what the type graph
+   says.  It is what makes the exemption un-sprinklable — a fixer agent cannot
+   escape the rule by adding a docblock line, because an invented service
+   method has no supertype and there is no legitimate reason to keep an unused
+   parameter on one.  You would simply delete it.
+
+2. **The author marked THAT parameter unused.**  The method docblock's
+   ``@param`` line for the flagged parameter carries an explicit
+   unused / ignored marker, e.g.::
+
+       @param string $userId Current user UID (unused)
+
+   Condition 1 alone would exempt a genuinely gutted implementation: if
+   ``RoleGuard`` stopped consulting ``$userId`` tomorrow that is a real defect,
+   and it is contract-imposed too.  The marker is what separates "deliberate"
+   from "forgot", and it is per-parameter, which
+   ``@SuppressWarnings(PHPMD.UnusedFormalParameter)`` is not — that annotation
+   is method-wide and cannot name which parameter it excuses.  #339 offered
+   both and preferred the PHPMD tag; the per-``@param`` marker is the stricter
+   of the two, and procest already carries it on all three findings while
+   ``MandaatGuard`` — which carries the PHPMD tag for a DIFFERENT parameter —
+   does not.  Keying on the PHPMD tag would have exempted MandaatGuard's
+   ``$userId`` for a reason that was never about ``$userId``.
+
+FAIL-CLOSED
+-----------
+Anything unresolvable is NOT an exemption: a supertype that lives outside this
+repository (an OCP interface, a vendored contract) cannot be inspected, so the
+finding stands.  Every error path exits non-zero.  A gate helper that cannot
+read its input must not answer "exempt".
+
+Usage:
+    check_caller_identity_exempt.py --root DIR --file F --line N \\
+        --method NAME --param '$userId'
+
+Exit codes:
+    0 — EXEMPT: contract-imposed AND explicitly marked unused
+    1 — not exempt (report the finding)
+    2 — could not decide (report the finding)
+"""
+
+import argparse
+import os
+import re
+import sys
+
+# `final class X`, `abstract class X`, `readonly final class X`, `interface X`,
+# `trait X` — capture the kind and the name.
+TYPE_DECL = re.compile(
+    r'^\s*(?:(?:final|abstract|readonly)\s+)*(class|interface|trait)\s+(\w+)\b'
+)
+UNUSED_MARKER = re.compile(r'\b(?:unused|ignored|not\s+used|no[t]?\s+consulted)\b', re.I)
+
+MAX_DEPTH = 6  # supertype BFS bound; a cycle guard already prevents loops.
+
+
+def read_lines(path):
+    with open(path, 'r', encoding='utf-8', errors='replace') as fh:
+        return fh.read().split('\n')
+
+
+def docblock_above(lines, idx):
+    """Return the docblock lines immediately preceding 0-based line `idx`.
+
+    Skips blank lines and PHP 8 attribute lines (`#[NoAdminRequired]`), which
+    legitimately sit between a docblock and its signature. Returns [] when the
+    preceding non-skippable line is not a docblock terminator.
+    """
+    i = idx - 1
+    while i >= 0 and (lines[i].strip() == '' or lines[i].strip().startswith('#[')):
+        i -= 1
+    if i < 0 or not lines[i].strip().endswith('*/'):
+        return []
+    end = i
+    while i >= 0 and '/**' not in lines[i]:
+        i -= 1
+    if i < 0:
+        return []
+    return lines[i:end + 1]
+
+
+def param_marked_unused(doc, param):
+    """True when the @param line for `param` carries an explicit unused marker."""
+    # `@param string $userId ...`, `@param $userId ...`, `@param string|null $userId ...`
+    pat = re.compile(r'@param\b[^$]*' + re.escape(param) + r'\b(.*)$')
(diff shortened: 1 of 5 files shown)


rubenvdlinde:
CI settled: **17 success, 17 skipped, 0 failure** (counted from the full check-run set via the API, not `gh pr checks`, which returns a partial set). `Package invariants` — the job that runs all 86 helper suites — is green.

One more measurement since opening, on the §1 recommendation in the #340 comment (delete the eight menu nodes from procest's own manifest sources, keep the pages, empty `removals`): a **full 60-gate run** on the modified tree moves exactly one verdict line, `[gate-53] effective-manifest-crossref: FAIL → PASS`, runner exit 7 → 6. Nothing else in the suite reacts, and the effective manifest is byte-identical. So that fix is behaviour-neutral against the whole gate suite, not just against gate-53.

Still **not merged** — the gate-53 half needs the decision in the #340 comment.

rubenvdlinde:
Correction to my earlier note, since it is now merged and the record should be right.

I reported the local `run-helper-suites.sh` run as "truncated at ~60/86, all reached suites green". That was wrong. A third attempt completed and the honest record is **20 failed + 2 quarantine-integrity failures**:

- **19** were `cd: .../hydra-gates: No such file or directory` — **self-inflicted**. I ran `rm -rf` on the working tree while the harness was still running in the background.
- **2** quarantine-integrity failures, same cause (discovery ran over the deleted tree).
- **1** genuine assertion failure that ran *before* the deletion, so it was not explained by it: `test_gate_crashed_checker_is_not_a_finding.sh` — *"gate-6 orphan-auth: emitted NO verdict line at all, though this fixture gives it a subject"*.

I A/B'd that one on a fresh clone rather than assume it was noise:

| arm | result |
|---|---|
| merged `main` `5523801` (**with** this change) | ALL PASS (run twice) |
| pre-merge `main` `c26f9a3` (**without** this change) | ALL PASS |

Three clean-tree runs across both sides of the change, all green — plus `Package invariants`, the job that sweeps all 86 suites, green on this PR's head. **The failure does not reproduce and is not attributable to this change.** This box was running four other agents with the disk at 96–97%, and that suite spawns the full 60-gate runner per assertion.

Nothing about the gate-3 or gate-53 conclusions changes. But the coverage claim for this PR rests on **CI's `Package invariants` job**, not on my local run — my local run inspected less than it appeared to, and then I deleted the evidence.

⚠️ Still open regardless of the merge: the **gate-53 decision** in the #340 comment. The merge shipped the id-vs-route detection fix; it did not answer whether `removals-invariant` should be amended, downgraded to WARN, or rebuilt as a base/head effective-menu diff. procest's 8 are still red and still closable app-side.

